PR #3, opened 2022-05-06
Upgrade vue cli and node

jorntx:
Upgraded packages (including vue-cli) to match node 16.5.0. This major LTS node version will be the project default from now.


jorntx:
Conflicting package-lock.json

PR #8, opened 2022-08-02
Wok 356 search input implementation

jenskaj:
Added fonts, assets and styles along with search input from Pattern Labs


src/main.ts
```diff
@@ -1,4 +1,5 @@
 import { createApp } from "vue"
+import './assets/styles/style.scss'
```
jorntx, line 2: Move this import below the JS imports

jorntx (commented):
Fine first stab. We need to trim the imported SCSS and the hero images for a start. This can be done in separate PR's. 

I have one tiny tiny suggestion for improvement and then it's good to go

jorntx (approved):
Good to go

PR #10, opened 2022-08-03
Wok 329 implement full record for test records

jorntx:
Adds basic skeleton for full record view. Note that this is made for "mock data" så I have hardcoded the record id and img src url for now to enable easy testing.

Pay no attention to lack of styling, presentation and the metadata/img present or the precise content of the ts-types. Focus on the internal wiring/structure and the overall typings. If we are happy with the direction of this commit we will polish the before mentioned in the next iterations.

NOTE: please review https://github.com/kb-dk/ds-web/pull/9 before this one as it will make merging easier:)


src/components/records/recordItem.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+	<div class="">
+		<div>Author: {{ recordItemData.author[0].name['@value'] }}</div>
```
jenskaj, line 3: sure it's not `name[0]` like below?

jenskaj (approved):

src/components/records/recordItem.vue
jorntx: Yes, for some reason this differs in the Author data-block. I might add that the fetching the first entry of the array is hackish and temporary - when we get the final data we will of course list all authors, subjects, etc.

PR #12, opened 2022-08-04
WOK-362: roughly sort assets and stylesheet imports

jenskaj:


src/assets/styles/style.scss
```diff
@@ -16,169 +16,153 @@
...
-@import "atoms/logo";
-@import "atoms/switch";
-@import "atoms/input-number";
-@import "vendors/choices";
-
+// @import "atoms/dividers";
+@import 'atoms/buttons';
+@import 'atoms/burger-button';
+@import 'atoms/inputs';
+// @import "atoms/checkbox-and-radios";
```
jorntx, line 95: Looking at the design material we have I think we might need this one for the facets/filters.

jorntx (approved):
One small comment about an import I think we might need (still approved though). Other than that we are good to go and have a good first trim of the styles.

FYI unfortunately the bundler is still throwing warnings about the bundle size.

jorntx (approved):
Good to go

jorntx:
We need new PR to develop - I will close this one.

PR #14, opened 2022-08-11
Wok 361 initial search bar styling

jenskaj:


src/App.vue
```diff
@@ -1,15 +1,33 @@
...
+			>
+				<router-link to="/">Home</router-link>
+				|
+				<router-link to="/about">About</router-link>
+			</div>
+			<SearchBar />
+		</div>
+	</div>
+	<div class="content">
+		<router-view />
```
jorntx, line 16: If this continues to grow we should consider some component structure here. I like my App.vue nice, lean and clean:)

src/App.vue
```diff
@@ -1,15 +1,33 @@
...
+import { defineComponent } from 'vue';
+import SearchBar from './components/search/SearchBar.vue';
+
+export default defineComponent({
+	name: 'App',
+	components: { SearchBar },
+});
+</script>
+
+<style lang="scss">
```
jorntx, line 30: Good stuff with SearchBar component

src/components/search/SearchBar.vue
```diff
@@ -0,0 +1,102 @@
...
+	},
+});
+</script>
+
+<style
+	lang="scss"
+	scoped
+>
+@import './search.scss';
+</style>
```
jorntx, line 102: This works very well and for large amounts of component specific scss I think it will be excellent in terms of keeping the component manageable.

vue.config.js
```diff
@@ -24,4 +24,16 @@ module.exports = {
...
+		  sass: {
+			additionalData: `
+				@import "bootstrap/scss/functions.scss";
+				@import "bootstrap/scss/mixins.scss";
+				@import "bootstrap/scss/variables.scss";
+				@import "@/assets/styles/generic/custom-variables.scss";
+			`
+		  }
+		}
+	  }
```
jorntx, line 38: This turned beautifully!

jorntx (approved):
Approved:) - Just note the gotchas in terms of your query param refactoring (not the $router.push stuff but the $watch => $route stuff)

src/views/Search.vue
```diff
@@ -20,24 +17,28 @@ import { defineComponent } from 'vue';
...
+	created() {
+		// Watch query params and update search results if anything changes 
+		this.$watch(
+			() => this.$route.query.term,
+			(newTerm: string, prevTerm: string) => {
+				if (newTerm !== prevTerm) {
+					this.searchResultStore.getSearchResults(newTerm as string);
+				}
+			},
+		);
```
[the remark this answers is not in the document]
jenskaj: Isn't this the desired outcome though? We don't want to trigger anything if the route stays the same, right?

src/App.vue
jenskaj: I absolutely agree! I'll do that asap

src/components/search/SearchBar.vue
jenskaj: Yeah, I think we're on the right track with the styling setup

src/views/Search.vue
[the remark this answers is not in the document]
jorntx: Yes, this is the desired outcome in this case - so the comment was more fyi - as  you can run into cases where some logic is strapped to lifecycle hooks and combined with e.g. the use of segment paths you can run into the 'gotcha' where they are not triggered although a different user, record, item, query etc. is loaded.

PR #22, opened 2023-08-21
Disc 66/top level architecture

jesperlauridsen:
This is the top level architecture we talked about, this should be fairly straight forward. I'll do some JIRA issues on the missing pieces (like single record etc)


src/components/search/SearchResults.vue
```diff
@@ -1,14 +1,14 @@
...
 	<div
 		class="hit-box"
-		v-for="(res, i) in (currentResults as PropType<GenericSearchResult[]>)"
-		:key="res.id"
+		v-for="(res, index) in (currentResults as PropType<GenericSearchResult[]>)"
+		:key="res.id + '-' + index"
 	>
 		<kb-resultcomponent
+			:number="index"
 			:data="JSON.stringify(res)"
```
jorntx, line 9: Lets put the discussion about this on the board.

src/components/search/result-component.js
```diff
@@ -21,43 +21,48 @@ class ResultComponent extends HTMLElement {
...
 	attributeChangedCallback(name, oldValue, newValue) {
 		if (name === 'data') {
 			this.resultData = JSON.parse(newValue);
-			this.shadowRoot.querySelector('.title').textContent =
-				this.shadowRoot.querySelector('.title').textContent + ' ' + this.resultData.title;
+			this.shadowRoot.querySelector('.title').textContent = Number(this.number + 1) + ' ' + this.resultData.title;
+			this.shadowRoot.querySelector('.where').textContent = 'KANAL';
+			this.shadowRoot.querySelector('.when').textContent = 'TIDSPUNKT';
+			this.shadowRoot.querySelector('.duration').textContent = 'VARIGHED';
+			this.shadowRoot.querySelector('.summary').textContent =
```
jorntx, line 34: Lets put a ticket on the board to address translation.

src/components/search/result-component.js
```diff
@@ -21,43 +21,48 @@ class ResultComponent extends HTMLElement {
...
-			:src="imgSrc"
-			:alt="altTxt"
-		/>
-	</figure>
+			<img
+				class="image-item"
+				src="http://devel11.statsbiblioteket.dk:10001/ds-image/v1/IIP/?FIF=%2FDAMJP2%2FDAM%2FMaps%2F0000%2F067%2F807%2Fdk001991&ROT=0&CVT=jpeg"
+				alt="altTxt"
+			/>
+		</figure>
```
jorntx, line 65: I guess this is only temprorary until backend gets their ducks in a row

src/components/search/HitCount.vue
```diff
@@ -1,10 +1,10 @@
 <template>
-	<div>
+	<div v-if="query !== ''">
 		<div
 			class="hit-count"
 			v-if="hitCount > 0 || noHits"
 		>
-			Hits found - {{ hitCount }}
+			{{ hitCount }} resultater fundet for "{{ query }}"
```
jorntx, line 7: Again here we need to figure out how far we wish to go with translation and in this case in particular with pluralization

src/components/search/SearchResults.vue
```diff
@@ -1,14 +1,14 @@
 <template>
 	<div
 		class="hit-box"
-		v-for="(res, i) in (currentResults as PropType<GenericSearchResult[]>)"
-		:key="res.id"
+		v-for="(res, index) in (currentResults as PropType<GenericSearchResult[]>)"
+		:key="res.id + '-' + index"
```
jorntx, line 5: Need 'double uniqueness'?

src/components/search/search-component.js
```diff
@@ -4,6 +4,10 @@ class SearchComponent extends HTMLElement {
 		this.shadowRoot.innerHTML = TEMPLATE + STYLES;
 
 		this.searchQuery = this.shadowRoot.querySelector('#focusSearchInput');
+		if (location.search) {
+			let q = location.search.split('&')[0].split('=')[1];
```
jorntx, line 8: This works but we should look into some kind of utility method så we avoid the gotchas and it becomes easier to get location params

jorntx (approved):
Nice rearrangement and we will go with this for now. 

I have some general comments - mostly for reference - the main topic is that we should discuss how to handle translation.

I took the branch for a spin and noticed some lag on the rendering of search results. It is particularly noticeable when focusing on the hit count which appears instant and then a short wait before the animation of search results kicks in. 

I will merge as soon as you have read the comments.

PR #23, opened 2023-08-21
Disc 3/single result webcomponent

jesperlauridsen:
Simple styling of the single result (and the list of results). At the moment I don't see the point in doing any more on these, before we know more about that it's gonna be like (both on the structural side, as we're waiting for the radio/TV, and on the design side)

### DIFF
--- a/src/views/Search.vue
+++ b/src/views/Search.vue
@@ -1,13 +1,18 @@
 <template>
 	<div>
+		<div class="container">
 		<div class="hit-count">
 			<HitCount
 				:hit-count="searchResultStore.numFound"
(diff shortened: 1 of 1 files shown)


jorntx (approved):
I can see I should have reviewed the branches in reverse order:)

So the comments from DISC-66/top-level-architecture are also valid here so I will merge this right away. And I agree btw - lets cool down on detailed styling until we have better guidelines.

jorntx:
This PR is linked to DISC-3 but DISC-3 is already done and merged??

jorntx:
And I postponed merge till you are online again so we can minimise merge conflicts

PR #24, opened 2023-08-23
bump and lock to node 18.17.1 and npm 9.6.7

jorntx:
Bump and lock to node 18.17.1 and npm 9.6.7 both versions are lts.

@jesperlauridsen - just boot the app and see that everything works as before :)


jorntx:
Ufff, forgot to mention that I removed some unused packages as you can see in the change log..

jesperlauridsen:
Booted with no issues - all looks nice. Approving.

PR #25, opened 2023-08-23
Disc 72/delay bug

jesperlauridsen:
This should remove the delay bug when the results are the first on the page.


jorntx (approved):
Much better!

Took the liberty to merge master into this branch as they were out of sync.

PR #26, opened 2023-08-23
Disc 67/page transitions setup

jesperlauridsen:
A fairly straight forward page transition. We can make it simpler if we want - I really just wanted to make one that used a wipe effect.


jorntx (commented):
I like smooth transition! See suggestions for improvement in inline comments.

Things to consider when moving forward (mainly UX):
* Speed of the transition
* Is the transition to 'heavy' if users continuously go from search to detail view (postvisning) and back again and can we mitigate this?
* When we get the transitions right - we will have to document them so the rest of KB-systems kan implement them as needed because I guess we wouldn't want to ship transitions as part of the web component because app context would be needed.

jorntx (approved):
Good to go

PR #27, opened 2023-08-28
Disc 98/add facets to searchpage

jesperlauridsen:
This adds the first stab of facets to the search results. some minor things to be aware of:

Right now, adding a facet does a new search instant. This is how we've normally done this - I'm not sure if that's what we want going forward with these checkboxes - but for now, it adds the facets and makes it work.

Right now, the URL is not updated. I think this should be a seperate JIRA.


jorntx (changes requested):
Found some issues - mostly typings and some suggestions for better naming and shorter code.

src/components/search/facet-checkbox-component.js
jorntx, line 1: Lets remember a ticket so we remember to port this to .ts like the others

src/store/searchResults.ts
jorntx, line 1: Looks good with the typings!

src/types/APIResponses.ts
```diff
@@ -10,9 +10,18 @@ export interface APISearchResponse {
...
 }
 
+export interface Facets {
+	facet_queries: Array<string>;
+	facet_fields: Array<string>;
+	facet_ranges: Array<string>;
+	facet_interval: Array<string>;
+	facet_heatmaps: Array<string>;
+}
+
```
jorntx, line 24: Them backenders like it messy....

jorntx (approved):
Looks good - lets roll with this for now.

On the wish list - an abstracted solid 'facet engine' that can handle the book keeping around adding an removing facets.

PR #31, opened 2023-08-29
Disc 90/single record page

jorntx:
I think this is a first good stab at full record in regards to reaching M1. No styling of importance only basic scaffolding for metadata and video. I only implemented basic metadata for video and the video url is hardcoded so search for 'backstage' when testing :)

### DIFF
--- a/src/views/ShowRecord.vue
+++ b/src/views/ShowRecord.vue
@@ -1,12 +1,13 @@
 <template>
-	<!-- TODO handle empty response scenario -->
-
-	<div v-if="recordData">
-		<div v-if="recordType === 'BroadcastEvent'">
-			<BroadcastRecordMetadataView :record-data="(recordData as BroadcastRecord)" />
-		</div>
-		<div v-else>
-			<GenericRecordMetadataView :record-data="(recordData as GenericRecord)" />
+	<div>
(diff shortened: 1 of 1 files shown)


src/types/GenericRecord.ts
```diff
@@ -1,74 +1,67 @@
-// Expand as we go
 export interface GenericRecord {
+	'@context': Array<ContextClass | string>;
+	'@type': string;
 	id: string;
 	url: string;
+	'kb:admin_data': KBAdminData;
```
jesperlauridsen, line 6: Why is this in ' '?

src/views/ShowRecord.vue
```diff
@@ -1,39 +1,59 @@
...
-	<div>
-		<div
-			class=""
-			v-for="rec in recordStore.record"
-			:key="rec.id"
-		>
-			<RecordItem :record-item-data="rec" />
+	<!-- TODO handle empty response scenario -->
+
+	<div v-if="recordData">
```
jesperlauridsen, line 4: I was under the impression that vue wants everything in a template wrapped in a single div - so that you should have 2 divs inside with v-if="recordData" and v-else, but not on the top level?

PR #33, opened 2023-08-31
should fix the problem for now

jesperlauridsen:
This should fix the problem as of now - we still need to figure out a better way not to reboot the app from web components in M2.


jorntx (approved):
I agree something fishy is going on....

PR #35, opened 2023-08-31
Weird fade in/out bug on search results

jesperlauridsen:
Not really sure where this comes from. I swear it worked as intended before. This fixes it for now. Then we can change it when we have a better idea of the animation we want.


jorntx (approved):
Not shure what the problem was but it works fine now.

PR #36, opened 2023-08-31
Disc 109/build header and menu

jesperlauridsen:
First stab:

Only a desktop nav bar - needs to be build as a mobile menu aswell
hardcoded image for now on the search field. Alot of stuff there needs to be adjusted when Daniel has time to refine the design.

I think it works for now - to make it look a little more polished and to use as a talking point with UX/Design team.


jorntx (approved):
As we talked about we will go with this for now as a discussion point. Soooo much could change from now till production so review wise i justed checked it out, gave it quick spin and merged some other branches in.

PR #37, opened 2023-09-01
Disc 79/translation foundation

jorntx:
Adds basic foundation for translation with a couple examples


jorntx:
After the review you will have to 'merge pull request', rebase is not an option although I have merged everithing into this branch as the git log shows...

I need to strengthen my git mojo.

src/components/nav/menu-component.ts
```diff
@@ -5,27 +5,30 @@ class MenuComponent extends HTMLElement {
...
+		return ['locale'];
 	}
+
 	attributeChangedCallback(name: string, oldValue: string, newValue: string) {
-		if (name === '') {
-			//console.log("something")
+		if (name === 'locale') {
+			const localeSwitcher = this.shadow.querySelector('#localeSwitcher');
+			const localeSwitcherTxt = newValue === 'da' ? 'In English' : 'På dansk';
+			localeSwitcher && (localeSwitcher.textContent = localeSwitcherTxt);
```
jesperlauridsen, line 24: Perhaps this is not the most ideal placement of the text. When we have more things to be translated, we should probably collect everything in a translation object like the template and styles strings, so everything is neatly together.  That would also ensure that we know where translations are in every web component we build in the first step at least, making it easier to navigate and change as needed.

I'll let it slide for now - but we should create a JIRA on this issue and do it in all the web components as we begin to have components with actual text in them.

src/components/search/HitCount.vue
```diff
@@ -4,7 +4,7 @@
 			class="hit-count"
 			v-if="hitCount > 0 || noHits"
 		>
-			{{ hitCount }} resultater fundet for "{{ query }}"
+			{{ $tc('search.hitcount', hitCount) }} "{{ query }}"
```
jesperlauridsen, line 7: Neat!

src/components/nav/menu-component.ts
jorntx: I totally agree!

src/App.vue
```diff
@@ -33,9 +33,19 @@ export default defineComponent({
 		return {
 			td: 0.35,
 			leaveDone: false,
+			locale: 'da',
 		};
 	},
+	created: function () {
+		//Remember to check for init locale once we now where to get it from
+		window.addEventListener('locale-switch', this.switchLocale);
```
jesperlauridsen, line 41: I'm 99% sure this is fine and we don't need to worry about removing because it's done on dismount anyway by the browser. 99%.

jesperlauridsen (approved):
Everything looks very nice and tidy. I like it! My few comments are just general stuff, nothing that stops the commit. I'll approve and merge.

src/App.vue
jorntx: Nice catch! - yes I think so too but good to keep an eye on these event handlers.

PR #38, opened 2023-09-04
Add dynamic thumb image src url

jorntx:
Removes hardcoded image url and replaces with dynamic url from data response


jesperlauridsen (approved):
Nice.

PR #39, opened 2023-09-05
smooth image load, icons and css polish

jesperlauridsen:
Added a simple image load with a gradient showing untill image is ready
Added a few material icons to the search result
did some zindex cleaning and general polish


jorntx (approved):
Yes definitely gave it i nice brush up and the small icons work magic.

PR #40, opened 2023-09-05
This should make it so routing is done by the app (and made optional)

jesperlauridsen:
I think this works well (hopefully you agree!)

### DIFF
--- a/src/components/search/result-component.ts
+++ b/src/components/search/result-component.ts
@@ -1,7 +1,7 @@
 class ResultComponent extends HTMLElement {
 	shadow: ShadowRoot;
 	number: number | undefined;
-
+	vueRouting: boolean | undefined;
 	constructor() {
 		super();
 		this.shadow = this.attachShadow({ mode: 'open' });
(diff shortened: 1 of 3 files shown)


src/components/search/result-component.ts
jorntx, line 1: I like this solution and it works. The app. now handles the routing.

src/components/search/result-component.ts
```diff
@@ -22,14 +22,24 @@ class ResultComponent extends HTMLElement {
 	}
 
 	static get observedAttributes() {
-		return ['data', 'number', 'show'];
+		return ['data', 'number', 'show', 'vueRouting'];
```
jorntx, line 25: I really wish there was a way around this. I takes a small toll each time we full time observe properties. In the case of vueRouting this will likely never change once the web component is bootstrapped.

src/components/search/result-component.ts
```diff
@@ -60,14 +70,16 @@ class ResultComponent extends HTMLElement {
...
+		}
 	}
 }
 
 const RESULT_COMPONENT_TEMPLATE = /*html*/ `
 	<div class="container">
 		<div class="information">
-		<a href="" class="title"></a>
-		<div class="subtitle"><span class="where"></span><span class="when"></span><span class="duration"></span></div>
+		<a role="link" class="title"></a>
```
jorntx, line 82: With a few more enrichments I think this will work nicely with the  requirements.

jorntx (approved):
Looks good and solves a big issue \o/

jorntx:
Had to resolve some merge conflicts before I could merge to master

PR #41, opened 2023-09-05
Should fix the issue with result sizes and weird scrollbars on the x-axis

jesperlauridsen:
Just some css cleanup


jesperlauridsen:
FYI - this also fixes the search field collapsing when there's a search result present.

jorntx (approved):
Again very nice touch up. Looks better every time you tighten up the UI:)

PR #42, opened 2023-09-05
rebuild header with translations

jesperlauridsen:
Now, a few notes to this PR:

I thought about making some small menu-element components, but decided against it for now because of the need for eventlisteners to click on certain elements in the menu. Bouncing events around seems unnessecary when I could achive what I wanted in the same component, without too much hassle or the component being too big.

Also, the menu rebuilds when changing - I could've made a separate function to take care of changing just translations, but that would just add more code to the file - with a minimal win (at least when I was looking it at it today)

Any feedback on this is greatly appreciated. It is, as we discussed, mainly a try to figure out how to proceed with translations in web components. So any input is nice here!

### DIFF
--- a/src/components/nav/menu-component.ts
+++ b/src/components/nav/menu-component.ts
@@ -1,25 +1,7 @@
-type TranslationItem = {
-	title: string;
-	link: string;
-	icon?: string;
-	id?: string;
-};
-
-type Translation = {
-	da: {
-		primary: TranslationItem[];
-		secondary: TranslationItem[];
-	};
-	en: {
-		primary: TranslationItem[];
-		secondary: TranslationItem[];
-	};
-};
-
 class MenuComponent extends HTMLElement {
 	shadow: ShadowRoot;
 	lang: string;
-	translation: Translation;
+	translation: MenuTranslation;
 
 	constructor() {
 		super();
@@ -129,6 +111,24 @@ class MenuComponent extends HTMLElement {
 		parent.appendChild(listElem);
 	}
 }
+
+type MenuTranslationItem = {
+	title: string;
+	link: string;
+	icon?: string;
+	id?: string;
+};
+
+type MenuTranslation = {
+	da: {
+		primary: MenuTranslationItem[];
+		secondary: MenuTranslationItem[];
+	};
+	en: {
+		primary: MenuTranslationItem[];
+		secondary: MenuTranslationItem[];
+	};
+};
 
 const MENU_COMPONENT_TEMPLATE = /*html*/ `
 	<header
--- a/src/components/nav/menu-component.ts
+++ b/src/components/nav/menu-component.ts
@@ -1,16 +1,73 @@
+type TranslationItem = {
+	title: string;
+	link: string;
+	icon?: string;
+	id?: string;
+};
+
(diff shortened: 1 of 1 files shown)


src/components/nav/menu-component.ts
```diff
@@ -1,16 +1,73 @@
...
+	da: {
+		primary: TranslationItem[];
+		secondary: TranslationItem[];
+	};
+	en: {
+		primary: TranslationItem[];
+		secondary: TranslationItem[];
+	};
+};
+
```
jorntx, line 18: If it is all the same to you I would like to move this down (just before the HTML). I like it when the class declaration and constructor is right up in ya' face when looking through components.

src/components/nav/menu-component.ts
```diff
@@ -19,16 +76,58 @@ class MenuComponent extends HTMLElement {
...
+		if (main) {
+			main.innerHTML = '';
+			this.createSingleMenu(main, this.translation[this.lang === 'da' ? 'da' : 'en'].primary, 'rdl-primary-nav');
+			this.createSingleMenu(main, this.translation[this.lang === 'da' ? 'da' : 'en'].secondary, 'rdl-secondary-nav');
+			const localeSwitcher: HTMLAnchorElement | null = this.shadow.querySelector('#localeSwitcher');
+			localeSwitcher
+				? localeSwitcher.addEventListener('click', (e) => {
+						this.dispatchLocaleSwitch(e);
+				  })
+				: null;
```
jorntx, line 100: Somehow this recreation of the menu doesn't stack in terms of double elements or events, so the web component/browser must somehow clean up on re-render.

src/components/nav/menu-component.ts
```diff
@@ -19,16 +76,58 @@ class MenuComponent extends HTMLElement {
...
 		window.dispatchEvent(new Event('locale-switch'));
 		e.preventDefault();
 	}
+
+	createFullHeaderMenu() {
+		const main: HTMLElement | null = this.shadow.querySelector('.rdl-main-navigation');
+		if (main) {
+			main.innerHTML = '';
+			this.createSingleMenu(main, this.translation[this.lang === 'da' ? 'da' : 'en'].primary, 'rdl-primary-nav');
+			this.createSingleMenu(main, this.translation[this.lang === 'da' ? 'da' : 'en'].secondary, 'rdl-secondary-nav');
```
jorntx, line 94: I will allow it this time but that call to createSingleMenu with the snug ternary providing the locale is balancing on a sharp edge in terms of readability:)

jorntx (commented):
I think this is a very good start. The built in types may balloon as we go and force us to externalize them but still keep them bundled with the web component.

I have only one suggestion for improvement in the comments.

src/components/nav/menu-component.ts
jesperlauridsen: main.innerHTML = ''; takes care of this - i remove everything from the div just before i add something new.

src/components/nav/menu-component.ts
jesperlauridsen: Good idea, will do.

jorntx (approved):
Looks good!

PR #43, opened 2023-09-06
Disc 118/global error handling

jorntx:
This took a really long time and quite a few iterations. Take a look at it and then let's talk about if this is the right approach.

I built a simple error-manager (which can be tuned to our needs) and I injected it globally in the app through Vues provider hooks. 

I tapped into request/response Axios interceptors for future use - they are not ciritical for the core of the error-manager right now. 

Ohhh and I forgot to mention a very important thing - I had to refactor the search store so it now uses a more setup style syntaks (mentioned as an option in the Pinia docs). I needed to do this in order to use the 'inject' framework.


src/error-handling/ErrorManagerType.ts
```diff
@@ -0,0 +1,5 @@
+import { AxiosError } from 'axios';
```
jesperlauridsen, line 1: This type should probably be moved to live with the other types we have in the project ;-)

src/error-handling/error-manager.ts
```diff
@@ -0,0 +1,78 @@
...
+		return { type: 'error', message: 'WE NEED SOME GOOD ERROR KEYS HERE' };
+	}
+
+	public submitError(error: AxiosError): void {
+		// Check if the current error is in the error history
+		if (!this.isErrorInHistory(error)) {
+			this.publishNotifyEvent(error);
+			this.errorHistory.push(error);
+
+			// Remove error after 5 seconds
```
jesperlauridsen, line 57: Curious why this is needed. When the error is published, could we not just remove it right away? Whats the reason for the 5 second delay? Alternativly - should the notification module instead submit an event when the error has been displayed to remove it?

src/store/searchResults.ts
```diff
@@ -1,57 +1,71 @@
...
+	const errorManager = inject('errorManager') as ErrorManagerType;
+	const numFound = ref(0);
+	const loading = ref(false);
+	const error = ref('');
+	const currentQuery = ref('');
+	const noHits = ref(false);
+	const filters = ref([] as Array<string>);
 
-export const useSearchResultStore = defineStore({
-	id: 'searchResults',
```
jesperlauridsen, line 10: The id is gone in the new way the store is made - Is this an oversight, or are we just not using it for anything?

jesperlauridsen (changes requested):
This looks very solid - and super cool. I think it's a very good foundation for the notification/error system for the app.

I've asked a few questions in the comments - move the types and we're all good to merge! ;)

Good job!

src/error-handling/ErrorManagerType.ts
jorntx: Yes, totally agree and you supplied good arguments:)

src/error-handling/error-manager.ts
jorntx: Only to suppress the scenario that the _exact same_ error is submitted several times within a very short time frame. I don't think that multiple events of the exact same kind (within a short time frame - e.g. 3 seconds) are necessary to dispatch to the user.

I have no hook for call backs from the notifier so I think if the error-notification should be suppressed or removed internally in the error-manager it should be by some internal metric.

In regards to keeping the notifications relevant and dismissing them I will confidently outsource this to the notifier.

src/store/searchResults.ts
jorntx: It's definitely an oversight (good catch) and frankly I don't now if we use it all. To me the app functions as normal.

jorntx:
Types moved as requested

jesperlauridsen (approved):
I think this is an excellent foundation for the error handling. Well done!

PR #44, opened 2023-09-08
Move result data from observed attribute to property in web component

jorntx:
So another deep dive into the docs of Vue3 and web components. This is the first draft on how we can do this without the heavy use of observables (when not needed). I have passed the complex search result data as a property, declared it is a property in the web component and then made use of the standard connectedCallback to render the results. This approach seems to work and reduce the "observer load"

Learnings when reading through the docs:
 - we should not in (nearly) any case pass complex data as attributes (only simple data). Complex data should be passed as a property and it turns out if your do that _and_ you have declared the property in the web component Vue 3 uses the 'in' operator to check for the presence of DOM properties and prefers to set the value as a DOM property if the key is present - pure magic! This will work most of the time but there are some exceptions (hopefully we will know when we meet them :) )

Take a look at it and if you like what you see I will adapt to this approach going forward (and refactor) :))

### DIFF
--- a/src/components/nav/menu-component.ts
+++ b/src/components/nav/menu-component.ts
@@ -83,7 +83,7 @@ class MenuComponent extends HTMLElement {
 		}
 	}
 
-	createSingleMenu(parent: HTMLElement, menu: Array<TranslationItem>, name: string) {
+	createSingleMenu(parent: HTMLElement, menu: Array<MenuTranslationItem>, name: string) {
 		const currentMenu = document.createElement('ul');
 		currentMenu.setAttribute('role', 'menubar');
 		currentMenu.classList.add(name);
--- a/src/components/search/result-component.ts
+++ b/src/components/search/result-component.ts
@@ -1,7 +1,11 @@
+import { GenericSearchResult } from '@/types/GenericSearchResult';
+
 class ResultComponent extends HTMLElement {
 	shadow: ShadowRoot;
 	number: number | undefined;
 	vueRouting: boolean | undefined;
+	resultData: GenericSearchResult | undefined;
+
 	constructor() {
 		super();
 		this.shadow = this.attachShadow({ mode: 'open' });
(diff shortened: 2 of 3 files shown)


src/components/nav/menu-component.ts
```diff
@@ -83,7 +83,7 @@ class MenuComponent extends HTMLElement {
 		}
 	}
 
-	createSingleMenu(parent: HTMLElement, menu: Array<TranslationItem>, name: string) {
+	createSingleMenu(parent: HTMLElement, menu: Array<MenuTranslationItem>, name: string) {
```
jesperlauridsen, line 86: Yea, this should have been in the other one, by mad!

src/components/search/result-component.ts
```diff
@@ -28,47 +32,54 @@ class ResultComponent extends HTMLElement {
 	}
 
 	static get observedAttributes() {
-		return ['data', 'number', 'show', 'vueRouting'];
+		return ['number', 'show', 'vueRouting'];
```
jesperlauridsen, line 35: Could you not do the same with the Vue routing, and maybe even the number attributes aswell? They don't need to be dynamic either I believe?

jorntx: We could do this with a lot of them so as stated in the PR comment if you like what you see we will adapt this approach going forward and refactor.

jorntx: Didn't want to refactor upfront if you didn't like it :)

jesperlauridsen: I like it very much. Lets just merge this PR, and make a refactoring ticket in JIRA for the remaining parts. I think it's a very good way to reduce the load on the app with all these web components we're gonna use. Good find!

jesperlauridsen: Just hit this with a thumbs up if you agree then i'll merge it in :)

jesperlauridsen (approved):

PR #45, opened 2023-09-11
A simple spinner for now.

jesperlauridsen:


src/App.vue
```diff
@@ -1,5 +1,14 @@
...
 	<div class="container">
+		<transition name="fade">
+			<div
+				v-if="spinner"
+				class="spinner"
+			>
+				<div class="spinner-cube"></div>
+				<div class="spinner-cube-shadow"></div>
+			</div>
+		</transition>
```
jorntx, line 11: I like out new bouncy bouncy spinner :)))

src/App.vue
```diff
@@ -40,12 +50,22 @@ export default defineComponent({
 		//Remember to check for init locale once we now where to get it from
 		window.addEventListener('locale-switch', this.switchLocale);
 		window.addEventListener('change-path', this.gotoPath);
+		window.addEventListener('show-spinner', this.showSpinner);
+		window.addEventListener('hide-spinner', this.hideSpinner);
```
jorntx, line 54: I think we should go with a global store that holds the spinner state and toggle it from there. It seems more Vue-like, it gives us the possibility to have advanced 'toggle code' tucked away from App.vue and from your example of usage here it doesn't seem like we are going to trigger it a lot from web components. Even if that was the case we could manage that with a relay from the parent component to the store.

What is your opinion on this?

jorntx (commented):
I like the bouncy spinner - I really do. Hope we get to keep it:)

I have one general comment/suggestion I need you to consider/give feedback on

src/App.vue
jesperlauridsen: Hmm. Interesting thought. I initially went with this because it was the most lightweight way to do it. I do agree that, for the sake of code splitting and responsibility of functions, this could be its own component. A store makes sense if you wanna use Vue to trigger it. I'll see if i can do a nice rewrite to as store.

src/store/spinner.ts
jorntx, line 1: Your implementation is fine, textbook example and works like a charm.

I'm torn about us using two different syntax styles for the store. I like the old style (used here) but in some use cases this won't cut it. I could be convinced to live with diversion but then lets make it explicit - maybe add a comment in the 'deviant stores'.

jorntx (approved):
The refactor looks good. I have one comment for diskussion but nothing that stops the merge.

jorntx:
Merge master in and then I will see if I can rebase then

src/store/spinner.ts
jesperlauridsen: You're an old, wise man (;-)). I've refactored the store to be of the new type. Give it a quick look over and see if you're satisfied with it. I've merged in master and done some small cleanup of the PR. No rebase possible I'm afraid :|

jorntx, line 1: Nice refactor:)

jorntx (approved):
Looks good - thanks for the store refactor - lets get this puppy included into master

PR #46, opened 2023-09-11
Disc 91/better query update search comp

jorntx:
A proposal for better and safer query param handling. 

If we feel for it we can wrap the handling in a utility but I kind of think that the native API is clean enough and that we can live with the small caveats e.g. the native API handling multiple params (with same name) as an array etc.

### DIFF
--- a/src/components/search/search-component.ts
+++ b/src/components/search/search-component.ts
@@ -11,8 +11,11 @@ class SearchComponent extends HTMLElement {
 		const searchQuery: HTMLInputElement | null = this.shadow.querySelector('#focusSearchInput');
 		if (searchQuery) {
 			if (location.search) {
-				const q = location.search.split('&')[0].split('=')[1];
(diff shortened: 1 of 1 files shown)


jesperlauridsen:
I like this very much. I didn't know you could do that. I'm no solr query master, but doing a little testing with some more advanced queries, like:

(title:"Error 401" AND body:"Authorization is denied") OR title:Password

still got me the connect value, so I think it's good. The only thing in perhaps wondered is if we should get @tokee to hammer on it a little bit, see if he can break it in any way?

jorntx:
@tokee can hamme all he want as long as he follows the URL spec --> https://url.spec.whatwg.org

jesperlauridsen (approved):

jorntx:
I deployed to devel and @tokee broke it by ignoring the specification and searching for %gg.

I have created a seperate ticket for handling this "bug"

PR #47, opened 2023-09-12
First stab at a global notifications system

jesperlauridsen:
This is just a draft, since it has included some buttons for test purposes. If we deem this as a good start, i'll remove the helper stuff and convert to a real pull request.

The buttons work to show what happens if several errors drop after eachother. The global error handler is now connected, so manipulating the search or fullrecord api calls now throws an error that's picked up by the notificationmanager aswell.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,6 +1,6 @@
 <template>
 	<div class="container">
-		<NotificationManager></NotificationManager>
+		<Notifier></Notifier>
 		<kb-menu :locale="locale"></kb-menu>
 		<div
 			class="wipe"
@@ -27,12 +27,12 @@
 import { defineComponent } from 'vue';
 import './components/nav/menu-component';
 import gsap from 'gsap';
-import NotificationManager from '@/components/notification/NotificationManager.vue';
+import Notifier from '@/components/notification/Notifier.vue';
 
 export default defineComponent({
 	name: 'App',
 	components: {
-		NotificationManager,
+		Notifier,
 	},
 	data() {
 		return {
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,5 +1,6 @@
 <template>
 	<div class="container">
+		<NotificationManager></NotificationManager>
 		<kb-menu :locale="locale"></kb-menu>
 		<div
 			class="wipe"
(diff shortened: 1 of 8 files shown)


src/components/notification/NotificationManager.vue
```diff
@@ -0,0 +1,159 @@
...
+	setup() {
+		const notificationStore = useNotificationStore();
+		return { notificationStore };
+	},
+
+	created: function () {
+		window.addEventListener('notify-user', this.newNotification);
+	},
+	beforeUnmount() {
+		window.removeEventListener('notify-user', this.newNotification);
```
jorntx, line 58: I guess this is the way to go (instead of 100% store implementation) because of the support for web components?

src/store/notificationStore.ts
jorntx, line 1: I guess you are comfortable with the new store style already:)

jorntx (commented):
Looks good, I have only minor comments.

src/App.vue
```diff
@@ -1,5 +1,6 @@
 <template>
 	<div class="container">
+		<NotificationManager></NotificationManager>
```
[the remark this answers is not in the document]
jesperlauridsen: Good point, will do!

src/components/notification/NotificationItem.vue
```diff
@@ -0,0 +1,149 @@
...
+<script lang="ts">
+import { defineComponent } from 'vue';
+import { gsap } from 'gsap';
+import GSAPTween from 'gsap';
+
+export default defineComponent({
+	name: 'NotificationItem',
+	data: () => ({
+		duration: { time: 2, delay: 0 },
+		notificationAnimation: null as null | GSAPTween,
```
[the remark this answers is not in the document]
jesperlauridsen: As I said in our talk, this was the way (at least the one i could find! :)

src/components/notification/NotificationManager.vue
jesperlauridsen: Yes, but also because I couldn't find another decent way to talk to the error handler - the way it was made actually also communicated errors like this.

jorntx (approved):
Looks good - merging in!

jorntx:
Very nice job @jesperlauridsen

PR #48, opened 2023-09-14
Catch MalformedURIException when decoding query

jorntx:


src/components/search/search-component.ts
```diff
@@ -29,6 +25,30 @@ class SearchComponent extends HTMLElement {
...
+	private updateSearchQuery(searchQuery: HTMLInputElement) {
+		//Note - this does not handle arrays
+		try {
+			const q = new URL(location.href).searchParams.get('q');
+			if (q !== null) {
+				searchQuery.value = decodeURIComponent(q.toString());
+			}
+		} catch (error) {
+			if (error instanceof URIError) {
+				// TODO dispatch to errorManager og direct to Notifier
```
jesperlauridsen, line 37: Is there anything blocking us from implementing this already now? I think both the errorManager and the notifier has been implemented now? Would be a decent first test of the whole system, imho.

jesperlauridsen (approved):

PR #49, opened 2023-09-18
Disc 169/test kaltura embedding

jorntx:
I abandoned the provided embed code and tried to embed a bundled/compiled Kaltura player. This turned out to be a hazzle but take a look and let me hear what you think.


jesperlauridsen (approved):
I think this works excellent as a proof of concept for the use of a kaltura player. If we decide to go with it, there's some cleanup to do. But I don't think it's worth doing it yet. So i'll approve it for now.

jorntx:
I looked at the player error and it happens async in the Kaltura player, so there is no way of catching it outside their code except registering an error handler on the player. I don't think that is worth the hassle until we know exactly which player we are going with.

PR #51, opened 2023-09-21
Disc 195/duration bug

jorntx:
- Fixed duration bug
- Now duration handles hours as well as minutes and seconds
- Moved duration handling to separate vue component


src/components/records/components/Duration.vue
```diff
@@ -0,0 +1,57 @@
...
+					}
+					if (seconds > 0) {
+						durationParts.push(`${seconds}sek`);
+					}
+					formattedDuration.value = durationParts.join(' ');
+				} else {
+					console.log('No match found for the ISO duration format.');
+				}
+			} else {
+				console.log('No ISO duration provided.');
```
jesperlauridsen, line 43: Logging that no ISO duration was found is nice - but useless for users who doesn't check the console. Perhaps it should read "Ukendt varighed" / "Unknown duration" or something similar if there was no duration found?

jorntx: Good point and I agree. Actually I think we should suppress it all together because we will run into cases where we have start/end time but no duration. 

The gold plate solution would be to supply both start/end time to the duration component and if there is no precalculated duration we do the work ourself. If no duration AND no start or end time we write unknown duration.

jorntx: I went for the gold plate solution see comments below...:)

jorntx:
So I went for the gold plated solution, fixed the bug and added the missing translation and did some hardening of the duration parser. The full commit log:
 

- Restructure the duration methods to handle translations
- Add skeleton for unit tests
- Bump and changing dependencies and jest/babel/eslint
- Change jest/babel config
- Add unit test for various cases of bogues duration and start/end dates
- Disable Legacy mode for i18n so it works with composition API mode
- Restructered and hardened duration methods to make tests pass
- Add duration translations
- Change typing in BrodcastRecord
- Fix missing support for $tc in hitcount and notification

src/components/records/components/Duration.vue
```diff
@@ -0,0 +1,104 @@
...
+
+<script lang="ts">
+import { defineComponent, PropType, ref, onMounted, watch } from 'vue';
+import { useI18n } from 'vue-i18n';
+
+export default defineComponent({
+	name: 'Duration',
+	props: {
+		isoDuration: {
+			type: String as PropType<string>,
```
jesperlauridsen, line 13: What is the difference here, from just doing type:string? If I do type: String it still seems valid? What am I missing here?

src/components/records/components/Duration.vue
```diff
@@ -0,0 +1,104 @@
...
+			required: true,
+		},
+		endDate: {
+			type: String as PropType<string>,
+			required: true,
+		},
+	},
+
+	setup(props) {
```
jesperlauridsen, line 25: You've build everything into the setup function here - is there a reason for not having the methods set up in the methods part instead?

src/components/records/components/Duration.vue
```diff
@@ -0,0 +1,104 @@
...
+			formattedDuration.value = t('duration.unknown');
+		};
+
+		const formatDuration = () => {
+			const isoDuration = props.isoDuration;
+			const startDate = props.startDate;
+			const endDate = props.endDate;
+
+			if (isoDuration && isoDuration !== '') {
+				const match = isoDuration.match(/PT(\d+H)?(\d+M)?(\d+S)?/);
```
jesperlauridsen, line 57: Regex man. It *looks* solid!

jesperlauridsen (approved):
I have a few comments, but they are mostly of a curious nature. Approved ! :1st_place_medal: Well done with the goldplating! A new level has been achieved with the unit tests too!

src/components/records/components/Duration.vue
jorntx: Stackoverflow for the win!

src/components/records/components/Duration.vue
jorntx: The reason for building everything in the setup function is twofold here:

1. The composition API best practice is to have it in the setup function and then export the methods needed in the templates (Yes, I know we have inconsistency on this in our app :) )
2. In this case everything is handled on onMounted nothing is called from the template, so the methods naturally belong to the core of the component as "private" methods and this is setup() in the composition API. If it gets too messy visually methods can be moved out from setup() and just referenced/called in setup.

src/components/records/components/Duration.vue
jorntx: Very good catch! - I have been messing with the typescript validator but ended up using an older version instead.

In the version I upgraded to I got an error when using only String - but this version does not really care so actually you are right - `String` would work perfectly fine and would be more concise. The only reason for going down this road would be if you were using Typescript to på more concise --> `String as PropType<Duration> `

jorntx (commented):
Already merged - answers to comments in comments :)

PR #52, opened 2023-09-21
Disc 228/mobile menu

jesperlauridsen:
This should add the first stab at a mobile menu - this has taken some quirking around to make it fit in - and I'll revisit it again, when we have to add the extra player of links in the mobile menu, Im quite sure.

The file is getting really big, but this is mainly because of all the css styling, which I'm afraid is required from the way it's made in the current running version.


src/components/nav/menu-component.ts
```diff
@@ -1,10 +1,13 @@
...
 class MenuComponent extends HTMLElement {
 	shadow: ShadowRoot;
 	lang: string;
 	translation: MenuTranslation;
-
+	collapsed: boolean;
 	constructor() {
 		super();
+		this.collapsed = true;
 		this.translation = {
```
jorntx, line 11: Maybe move the translations down to the  `MENU_COMPONENT_TEMPLATE`, `MENU_COMPONMENT_STYLES` and menu typing stuff to compact the constructor?

src/components/nav/menu-component.ts
```diff
@@ -68,6 +75,34 @@ class MenuComponent extends HTMLElement {
 		e.preventDefault();
 	}
 
+	toggleMenu() {
+		//this.shadow.querySelector('.rdl-main-navigation-wrapper')?.classList.toggle('collapse');
```
jorntx, line 79: Unused disabled code?

src/components/nav/menu-component.ts
```diff
@@ -68,6 +75,34 @@ class MenuComponent extends HTMLElement {
...
+		this.shadow.querySelector('#mobileNavButton')?.setAttribute('aria-expanded', this.collapsed.toString());
+		const expand = this.shadow.querySelector('.rdl-main-navigation-wrapper');
```
jorntx, line 81: Read the `expand` const as a boolean due to its naming but it really is a DOM element so maybe a rename too `navigationWrapper`?

src/components/nav/menu-component.ts
```diff
@@ -355,41 +429,158 @@ const MENU_COMPONMENT_STYLES = /*css*/ `
...
+		}
+	}
+	/* MEDIA QUERY 640 */
+	@media (min-width: 640px) {
+		.container {
+			max-width: 990px;
+		}
+	}
+	/* MEDIA QUERY 990 */
 	@media (min-width: 990px) {
```
jorntx, line 463: I know this is prone for changing and that you want to do more but your rearrangement with the MEDIA QUERIES etc. already added some structure to this rather beefy css.

jorntx (approved):
Very good (and hard) work. I think it works fine for now. I tested it and it mimics kb.dk behaviour and design very well. I have only very minor suggestions for improvement.

One tiny detail I did notice with the UI is the weird white spacing between the header image and the header from the last break point and downwards. 

Once you press 'resolve' on the comments I will merge.

src/components/nav/menu-component.ts
jesperlauridsen: That was an excellent idea!

src/components/nav/menu-component.ts
jesperlauridsen: Yea, i reworked how to do this a couple of times :) removed now.

src/components/nav/menu-component.ts
jesperlauridsen: Valid point. Renamed.

src/components/nav/menu-component.ts
jesperlauridsen: :pray:

jesperlauridsen:
> One tiny detail I did notice with the UI is the weird white spacing between the header image and the header from the last break point and downwards.

I don't quite follow this. But perhaps we could merge this in, and make bug report on this with a screenshot?

src/components/nav/menu-component.ts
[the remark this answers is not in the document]
jesperlauridsen: :pray:

jesperlauridsen:
I've added the suggested changes and re-requested a reviews. Should be all good now :)

jesperlauridsen:
You can even rebase, wooo!

jorntx:
> > One tiny detail I did notice with the UI is the weird white spacing between the header image and the header from the last break point and downwards.
> 
> I don't quite follow this. But perhaps we could merge this in, and make bug report on this with a screenshot?

Good idea and definitely my bad for not supplying screenie:(

jorntx (approved):
Looks good!

PR #53, opened 2023-09-21
Disc 232/Static image as placeholder for no imges for the search results

jesperlauridsen:


jorntx (approved):
Very straight forward - works as intended.

PR #54, opened 2023-09-22
Disc 233/New background image for the searchbar - loaded from the local repo

jesperlauridsen:


src/assets/images/_Den_Sorte_Diamant-Laura_Stamer-min.jpg
jorntx, line 1: Definitely a better picture:) However I do notice that this bad boy adds 428kb to our bundle size:))

jorntx (approved):
Seems good and nice with a change of scenery. Had one comment about image size but I will merge and hope you see it in the notification stream.

PR #56, opened 2023-09-29
Add error handling for custom errors like malformedURI

jorntx:
This started out to be a small thing but I actually had to do a bit of refactoring to get it to work.

I had to refactor SearchBar.vue to support composition API and then extend error-manager to support new error type. The rest is just implementing the stuff where error handling occurs and adding translations (I certainly hope the UX crew will give the wording an overhaul). I also took the liberty to fix the typescript error in SerchResults.vue

### DIFF
--- a/src/error-handling/error-manager.ts
+++ b/src/error-handling/error-manager.ts
@@ -3,10 +3,11 @@ import { createApp } from 'vue';
 import { ErrorManagerType } from '@/types/ErrorManagerType';
 
 export class ErrorManager implements ErrorManagerType {
-	private errorHistory: AxiosError[] = [];
+	private errorHistory: (AxiosError | string)[] = [];
 
 	// Singleton pattern - we only want one instance
 	private static instance: ErrorManager | null = null;
+
 	public static getInstance(): ErrorManager {
 		if (!ErrorManager.instance) {
 			ErrorManager.instance = new ErrorManager();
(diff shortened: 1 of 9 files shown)


src/error-handling/error-manager.ts
```diff
@@ -43,30 +77,19 @@ export class ErrorManager implements ErrorManagerType {
...
-		if (!this.isErrorInHistory(error)) {
-			this.publishNotifyEvent(error);
-			this.errorHistory.push(error);
-
-			// Remove error after 5 seconds
-			// TODO lets build a more solid error purge method...
-			this.removeErrorFromHistory(error, 5000);
+	private getNotifierMessage(error: AxiosError | string, message: string) {
+		if (typeof error === 'string') {
+			return { type: 'error', message: message };
```
jesperlauridsen, line 82: Are these to supposed to be the same?

src/error-handling/error-manager.ts
```diff
@@ -43,30 +77,19 @@ export class ErrorManager implements ErrorManagerType {
...
+		} else {
+			return { type: 'error', message: message };
 		}
```
jesperlauridsen, line 85: Are these to supposed to be the same?

jorntx: Dohhhh! Thanks - I corrected this. As long as we are talking about the publish notify event flow there is no need to differentiate between error types

jesperlauridsen (approved):

PR #58, opened 2023-10-03
streamline types and update usage

jesperlauridsen:
Alright, so in my opinion this was straight forward:

I did some small renaming for the types. If the file contains 1 interface, it ends in 'Type'. 2 or more, it's 'Types' as ending for the file.

Also, every interface type we have, is now name Type at the end. This should clear up any confusion there might be about what they are in the code.

I think this is the best way to do it. This way we can group the different types together, so it's easier to navigate, we can import what we need across, and everything should be better structural, imho.

As always, have a look and tell me if this is what you had in mind too :-)


jorntx:
I like the overall direction and the TitlePascalCamelCaseNaming. 

One thought - if a type definition goes from containing one 'type' to containing more this will trigger a refactor of the file name throughout the app.  

I have done a little refactoring along your lines to try to limit the exported interfaces to the absolute minimum to see if that would help but we still have a few that exports multiple interfaces so I think we should have an in office discussion about my point above.

jesperlauridsen:
I looked this over. I don't necessarily think it's a bad thing with the Type / Types endings. I mean, if we need to add another type to an already existing file and change Type to Types for that file, it just means we have to refactor the usage of that file and the places it's used - Which isn't that much different what we're normally doing - and really would make one consider making a NEW file with the new Types, unless they're connected in some way. Which imho is a good thing, since we want the types to be a little split out after where they're used I guess?

jorntx:
> I looked this over, @jorntx. I don't necessarily think it's a bad thing with the Type / Types endings. I mean, if we need to add another type to an already existing file and change Type to Types for that file, it just means we have to refactor the usage of that file and the places it's used - Which isn't that much different what we're normally doing - and really would make one consider making a NEW file with the new Types, unless they're connected in some way. Which imho is a good thing, since we want the types to be a little split out after where they're used I guess?

I totally buy these arguments so lets keep the new naming and accept the minor refactor. It will be up to the reviewer to catch any bobooos (breaking of this convention) from the responsible developer:)

Did you have a chance to look over my other committed adjustments?

jesperlauridsen:
Yes I did, they look very nice. I didn't think of removing export from some of the internal ones. Good thinking! I think it's ready to merge!

PR #59, opened 2023-10-09
Disc 263/component structure in repo

jesperlauridsen:
First stab at a nicer structure. Feel free to go nuts here!


src/components/search/wc_facet-checkbox.ts
```diff
@@ -1,27 +1,40 @@
...
+					);
+			  })
+			: null;
 
 		const observer = new IntersectionObserver((entries) => {
 			entries.forEach((entry) => {
 				if (entry.isIntersecting) {
-					this.style.transitionDelay = this.slide / 50 + 's';
-					this.style.opacity = 1;
+					this.style.transitionDelay = this.slide || 0 / 50 + 's';
```
jorntx, line 36: Don't quite understand what is going on here. If `this.slide` is undefined the `transitionDelay` will be `0/50 + 's'` - isn't `0 / 50 + 's'` the same as just setting it to '0s'?

jorntx (commented):
I like the overall direction that this is heading - good refactor of the struture. As we talked about yesterday:

* I like the 'wc' prefix for web components and I vote for we stick to 'wc-' instead of 'wc_' so we let the prefix alone be the determining part and limit the different naming styles we have (we also kebab case for typescript classes)
* Lets keep the pascal case naming for the vue components for now.  
*

src/components/search/wc_facet-checkbox.ts
jesperlauridsen: Should be fixed now!

jorntx:
This branch now needs a merge update:)

jorntx (approved):
Looks good to go - give me a heads up after you merge main in and I will merge this.

PR #60, opened 2023-10-12
Added copyscript

jesperlauridsen:
Made it so notifications can either get a key to translate, or get a message to display (which is kinda cool imho)

Did a little cleanup to make room for the expanded type.


jorntx (changes requested):
Looks good and I can see that this triggered some modification to the notification system which I think you implemented well.

After this review I will submit a PR to this branch with code using the new clipboard API instead of the deprecated document.execCommand('copy'). The reason why I don't push to this PR is because I can see you did a lot of work accommodating the textArea in different browsers so I want you to look my modification over and see if I overlooked something in terms of browser compatibility. 

If my proposal is good, then merge it in and then I will merge this branch. And of cause if it is no good - hit reject :))

jesperlauridsen:
It was all good - I've approved and merged. All should be good to go now!

jorntx (approved):
Good to go!

PR #61, opened 2023-10-13
Disc 183/persist chosen locale

jorntx:
First stab at persisting locale. I choose a combo between url and localstorage so we don't have to carry the url-param around on every single page but still maintaining the possibility to link/bootstrap the app in a certain locale.

Note DISC-303 (not coded yet) will complete the functionality by spreading this approach app wide. 

So see if you like where this is heading and then I will carry on with DISC-303 in a separate PR


jorntx:
Just merged a rather bulky main into this branch

src/App.vue
```diff
@@ -38,83 +40,116 @@ export default defineComponent({
...
+
+		const switchLocale = (e: Event) => {
 			e.preventDefault();
-			this.$i18n.locale = this.locale = this.$i18n.locale === 'da' ? 'en' : 'da';
-		},
+			locale.value = currentLocale.value = locale.value === 'da' ? 'en' : 'da';
+			LocalStorageWrapper.set('locale', locale.value);
+			router.push({
+				query: {
+					locale: currentLocale.value,
```
jesperlauridsen, line 67: This would remove any actual query that had been done, and push you to the frontpage without a query set. Is this intentional?

jesperlauridsen (changes requested):
Took it for a spin. The local store wrapper seems to work nicely, The only problem I have right now is that changing the language seems to reset everything done before it. Imho that should not be the case - Hopefully you agree and it's fixable :-) Otherwise we'll have a discussion about it!

src/App.vue
jorntx: No! And thanks for spotting this - this is a boboo on my part and I have comitted a fix.

jesperlauridsen:
Very noice! I'll accept it and make little bug report for the small issue I found!

jesperlauridsen (approved):

PR #66, opened 2023-10-26
Persist filters in URL

jorntx:
A really basic (but hard) first go at persisting filters in the URL


src/components/search/wc-facet-checkbox.ts
```diff
@@ -23,6 +23,7 @@ class checkboxComponent extends HTMLElement {
 						new CustomEvent('filter-update', {
 							detail: {
 								filter: `fq=${this.key}:"${this.sendValue}"`,
+								rawFilter: `${this.key}:"${this.sendValue}"`,
```
jesperlauridsen, line 26: filter and rawFilter are kinda they same. Would it be simpler to make filter the same value as rawFilter now, and simply add 'fq=' where it's needed when used to apply the filters to the URL? I'm not sure, just asking questions here...

src/components/search/Facets.vue
```diff
@@ -82,11 +80,33 @@ export default defineComponent({
...
+				} else if (Array.isArray(routeQueries.fq)) {
+					routeQueries.fq.push(newFilter);
+				} else {
+					//This will only trigger if someone manipulates the url manually
+					routeQueries.fq = [routeQueries.fq, newFilter];
+				}
 				this.searchResultStore.addFilter(e.detail.filter);
 			} else {
+				const filterToRemove = encodeURIComponent(e.detail.rawFilter);
+				routeQueries.fq = routeQueries.fq.filter((item: string) => item !== filterToRemove);
```
jesperlauridsen, line 106: This is some very nice coding. I had to read that a few times before i was sure what this did exactly :D

src/components/search/wc-facet-checkbox.ts
jorntx: Valid point and I had the exact same idea but eventually got too lazy to do the refactor, so added the param.

I think your call is good and we should take the time to rectify - so I will make a separate ticket for this.

src/components/search/Facets.vue
jorntx: This is ES6 at it's finest but please do speak up if it gets too compact.

src/components/search/wc-facet-checkbox.ts
jorntx: Ticket made - DISC-349

jesperlauridsen (approved):

jesperlauridsen:
Seems good - I'll approve and merge!

PR #67, opened 2023-10-31
Persist locale in url after locale switch

jorntx:
Locale should now be persisted in URL after first locale switch.

On init we rely on the value stored in localstorage (if any) otherwise we default to 'da'.


jesperlauridsen (approved):
Took it for a spin. Looks very good - I couldn't break it when i tried at least :-) Nice with a fairly simple and elegant solution to this. Kudos! :+1:

PR #68, opened 2023-10-31
Stream from Wowza

jorntx:
Stream from wowza, and switched from hardcaded to dynamic video url taken from data.
I switched back to videoJS player because it is a cleaner solution than Kaltura (for now) when streaming from Wowza

Note - we do not differ between audio or video yet... :)


jesperlauridsen (approved):
I got nothing for this - looks ve ry nice!

PR #69, opened 2023-11-01
Disc 329/searchpage responsiv

jesperlauridsen:
Allright. This took way longer than expected - And some polish is still missing - but that's due to design flaws / missing decisions.

Made it so facets can be toggled on/off on desktop.
Made it so facets can be hidden/displayed fullscreen on mobile.
Made it so a facet applied on mobile closes the facet window, so you can see the new result.
Made it so we check the kind of screen size on mount, to initialize the facets in the right way.
Made the design for mobile facets somewhat ready.

Facets.vue rewritten as composition API.

Fixed a bug where the query in the searchfield was not removed when reset.
Fixed a bug where eventlisteners on the facets were not removed on unmount, causing some weird stuff to happen.

A JIRA will be made for the tablet design - which wasn't ready when this was started.


src/views/Search.vue
```diff
@@ -145,6 +169,23 @@ export default defineComponent({
 			}
 		});
 
+		if (router.currentRoute.value.query.q !== undefined) {
+			searchResultStore.getSearchResults(router.currentRoute.value.query.q as string);
```
jorntx, line 173: You could go with `route.query.q` here is you did in the above code

src/views/Search.vue
```diff
@@ -236,17 +303,63 @@ h3 {
...
+	width: 0px;
+	transition: all 0.25s cubic-bezier(0.455, 0.03, 0.515, 0.955) 0s;
+	margin-right: 0px;
+	opacity: 0;
+	min-width: 0px;
+	margin-bottom: 20px;
+	position: fixed;
+	width: 100% !important;
+	top: 0px;
+	left: 0px;
```
jorntx, line 339: Do all this `0px` serve as 'resets'? If yes, we could consider grouping them in the future (for readability)

jorntx (approved):
This is very good work and it works fine - both the refactoring, the bug-fix and the responsive part. I think this serves as a good base for addressing the 'known issues' we talked about (either reworking the facets og adjust them to be more smooth). The responsive layout is usable as is and a good starting point.

I left some very minor comments but the nothing to hold back the PR. Look them over and ping me afterwards and I will merge.

src/components/search/SearchBarWrapper.vue
```diff
@@ -36,9 +49,9 @@ export default defineComponent({
 			}
 		});
 
-		watch(searchResultStore, (newValue, oldValue) => {
+		watch(searchResultStore, (oldValue, newValue) => {
```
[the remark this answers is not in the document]
jesperlauridsen: Errrh. That's a mistake on my part, fixed! Good catch!

src/views/Search.vue
jesperlauridsen: Will fix!

PR #71, opened 2023-11-02
DISC-384/Made the filter button like the one in the design

jesperlauridsen:
Pretty straight forward


jorntx (approved):
That button got buttered up nicely \o/

PR #73, opened 2023-11-03
Temporary fix for search filters containing '&'

jorntx:


jesperlauridsen (approved):
I'll approve this fix - seems to work nicely, and we got a JIRA for a long-term solution.

PR #74, opened 2023-11-03
Disc 321/pagination first step

jorntx:
Finally got a little time to attend to this:)

I took a look at some plugins and other implementations and I created a very stripped down version of a pager. 

I made this a draft as there is no styling and the pager is not hooked up to the api. This is just to test if the pager-mechanism itself works and is what we are aiming for. 

For the reference the pager plugin I looked at was 75.6kb uncompressed. This code is now 7kb uncompressed, so if it does the trick (and even with styling and hooks) it has saved some bandwidth:)

Ohhh and don't worry - the beefee comments will be removed once you have looked it over.


jorntx:
Is we talked this is bumped from draft to PR and I added functionality for updated the URL and then we will hook this up to the API later as well as add styling.

I also made an init method in the pager to bootstrap the correct state on reload or direct linking

src/components/search/Pager.vue
```diff
@@ -0,0 +1,159 @@
+<template>
+	<div>
+		<button
+			@click="prevPage"
+			:disabled="currentPageRef === 1"
+		>
+			Previous
```
jesperlauridsen, line 7: Translations :D

src/components/search/Pager.vue
```diff
@@ -0,0 +1,159 @@
...
+			>
+				{{ pageNumber }}
+			</button>
+		</span>
+		<span>{{ currentPageRef }} / {{ totalPages }}</span>
+		<button
+			@click="nextPage"
+			:disabled="currentPageRef === totalPages"
+		>
+			Next
```
jesperlauridsen, line 27: Translations! :D

jesperlauridsen (changes requested):
I think this looks very, very nice! Only thing missing is translations on the buttons.

Also, just a little note to us - when doing a new search, start should be set to 0 again. Right now it doesn't, but that's fine since it's not hooked up yet. We just need to remember to do that :)

src/components/search/Pager.vue
```diff
@@ -0,0 +1,170 @@
...
+		<span>{{ currentPageRef }} / {{ totalPages }}</span>
+		<button
+			@click="nextPage"
+			:disabled="currentPageRef === totalPages"
+		>
+			<span
+				class="nextPage"
+				:title="$t('search.nextPage')"
+			>
+				&gt;
```
jesperlauridsen, line 37: Im sorry to be an ass here - but I think, from the design perspective, that we need to use the material icons for the arrows. Like such:

	<i class="material-icons " aria-hidden="true">chevron_right</i>
	
	with chevron_left and chevron_right instead, they fit the design better.

src/components/search/Pager.vue
```diff
@@ -0,0 +1,170 @@
+<template>
+	<div>
+		<button
+			@click="prevPage"
+			:disabled="currentPageRef === 1"
+		>
+			<span
+				:title="$t('search.previousPage')"
+				class="prevPage"
+			>
```
jesperlauridsen, line 10: Im sorry to be an ass here - but I think, from the design perspective, that we need to use the material icons for the arrows. Like such:

	<i class="material-icons " aria-hidden="true">chevron_right</i>
	
	with chevron_left and chevron_right instead, they fit the design better.

jorntx:
Thanks - material icons added

jesperlauridsen (approved):
Very nice!

PR #79, opened 2023-11-09
Fix wrong init of locale when plucked from url

jorntx:


jorntx:
Thanks for testing this and finding yet a reload bug - I think this last commit wil take of it:)

PR #80, opened 2023-11-10
small fix, to take care of the scroll problem when leaving search.

jesperlauridsen:
If you go to a fullpost (fx /record/ds.radiotv:oai:du:1819f20a-20e4-44bb-bb9b-5b87e013cb7e), on mobile devices (<800px),  you cannot scroll on the page because of class on the body, set while on the searchpage to make sure filters work properly. This PR removes this when moving away from the search.vue module.

### DIFF
--- a/src/views/Search.vue
+++ b/src/views/Search.vue
@@ -48,6 +48,7 @@
 								class="search-facets active"
 								ref="facetsContainer"
 							>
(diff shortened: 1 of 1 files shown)


jorntx (approved):
Works fine:)

PR #81, opened 2023-11-13
DISC-380/tablet layout

jesperlauridsen:
This adds the first stab at the tablet facets.

It's set to be active between 640px and 800px, we're still waiting for what UX deems the right breakpoints.

Bonus: A little styling for the scrollbar on webkit-devices.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -269,7 +269,7 @@ body {
 	overflow: hidden;
 }
 
-@media (min-width: 800px) {
+@media (min-width: 640px) {
 	.remove-body-scroll {
 		max-height: initial;
 		overflow: initial;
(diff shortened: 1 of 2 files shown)


jorntx (commented):
Good to get started on this! I found a couple of things:

1. The 900 break looks good except I get that weird white spacing at the top - see screenie

2. Do the filters at the 640 break look as intended? - see screenie

jesperlauridsen:
Hey. The weird white spacing is part of the design, that's not 100% implemented yet.

image

The other lower poart is missing that's should cut into the search-background.

And yes, the filters are supposed to look like this on the first stab at tablet:

image

Although I haven't added the greyout screen behind because I'm not friends with this, and Daniel said it's still just wireframes, so we're not sure it would make it. I've kept it simple, using the same 'swipe-in-from-the-left' effect, so it's kinda they same across platforms (which is how i kinda think it should be).

Are these answers good enough for you, or do you want some changes off of this? :-)

jorntx:
> Hey. The weird white spacing is part of the design, that's not 100% implemented yet.
> 
> image
> 
> The other lower poart is missing that's should cut into the search-background.
> 
> And yes, the filters are supposed to look like this on the first stab at tablet:
> 
> image
> 
> Although I haven't added the greyout screen behind because I'm not friends with this, and Daniel said it's still just wireframes, so we're not sure it would make it. I've kept it simple, using the same 'swipe-in-from-the-left' effect, so it's kinda they same across platforms (which is how i kinda think it should be).
> 
> Are these answers good enough for you, or do you want some changes off of this? :-)

Very good good answers:)

PR #84, opened 2023-11-15
Moar typing!

jesperlauridsen:
I think this should do the trick?


src/api/api-client.ts
```diff
@@ -28,7 +28,7 @@ export class APIServiceClient {
 
 	//Search and record methods
 	async getSearchResults(query: string, filters: string): Promise<APISearchResponseType> {
-		return await this.httpClient.get(`search/?q=${encodeURIComponent(query)}&q.op=OR&indent=true&facet=true${filters}`);
```
jorntx, line 31: Nice to get this straightened out

jorntx (approved):
Works!

PR #87, opened 2023-11-16
Disc 378/split update filters method

jorntx:
I only split the method up and dried it out a bit - so functionality should be EXACTLY as before:)


jesperlauridsen (approved):
This is much, much better. Way easier to read and understand. Well done!

PR #88, opened 2023-11-16
Disc 435/top stylng on mobile and tablet

jesperlauridsen:
This adds some styling to the top on both the search page and the fullpost page. Should be enough for now - I've sent it back to the designers for review, hopefully they'll get it shaped up even better than this (on the fullpost, search should be good)


jorntx (approved):
This took care of it (compared it to devel).

While testing this I found another styling bug not related to this PR - will submit ticket.

PR #89, opened 2023-11-17
DISC-444/scroll to the top of the page after new search results are in.

jesperlauridsen:
This is my best guess at how this can be done smoothly. Don't look at the first commit - that solution was brutal!


jorntx (approved):
Works beatifully in FF and Chrome.

First pass ignored:)

PR #91, opened 2023-11-17
DISC-449/firefox-bug

jesperlauridsen:
This should fix the bug with the angled bar in the top, so it looks right in both firefox and chrome


jorntx (approved):
Fix verified in FF - nice!

PR #93, opened 2023-11-20
Fix for app not being able to build to production

jorntx:


jesperlauridsen (approved):
I agree with your comment - Good job at getting this to work!

PR #94, opened 2023-11-21
First pass at basic audio record

jorntx:
Basic view of audio record.

Ignore the video player stuff this will be fixed in separate ticket/PR


src/components/records/BroadcastAudioRecord.vue
```diff
@@ -0,0 +1,359 @@
...
+			>
+				<span class="material-icons">chevron_left</span>
+				Til forsiden
+			</router-link>
+		</div>
+		<div class="extra-record-data">
+			<div class="accordion">
+				<kb-accordion
+					first="true"
+					:title="t('record.moreMetadataBroadcast')"
```
jesperlauridsen, line 66: This change should probably also be in the BroadcastVideoRecord. Also, if we want reactivity, we might want to watch the locale and make a re-render when the language is changed - that doesn't happen right now.

jesperlauridsen (changes requested):
I just left a little comment. The translation on the title in the accordian should probably also be in the Video template one - and then a little question about the reactivity.

src/components/records/BroadcastAudioRecord.vue
jorntx: Agreed

jorntx: As we just talked about - lets take a stern look at out locale handling/reactivity in a separate ticket.

jesperlauridsen (approved):

PR #95, opened 2023-11-24
Disc 477/add audio player radio

jorntx:
Wooooosh, this was a hard one - first stab at implementing audio player. Went with Vidstack and finally got the css working (don't ask...).

The record or player is not finished styling wise yet as we agreed on waiting till the UX crew has a chance to weigh in.

I will create e separate ticket for migrating the video revord to Vidstack too. 

If the radio record design has a huge overlap with the video record design I will collapse the two vue components into one after migrating the video part.


jesperlauridsen (approved):
This all looks good. I can see you're just importing the styling in the stylesheet. I wouldn't have thought about that way to do it in a million years. Nice find!

PR #96, opened 2023-11-24
Disc 498/migrate video player

jorntx:
First pass at migrating video from VideoJS to Vidstack.

We need to find some kind of substitution for aspect-ratio in the css. I tried with calculation the width and padding, but no dice. I left a comment instead:)


src/components/viewers/AudioVideo/VideoPlayer.vue
```diff
@@ -21,45 +32,36 @@ export default defineComponent({
...
+@import 'vidstack/player/styles/default/theme.css';
+@import 'vidstack/player/styles/default/layouts/video.css';
 .video-player {
 	background-color: black;
 	display: flex;
 	justify-content: center;
 	align-items: center;
 	width: 100%;
-	aspect-ratio: 4/2;
+	/*aspect-ratio disabled - messes with the controls but now player is too high...*/
```
jesperlauridsen, line 63: Having a

 .player {
	aspect-ratio: 4/2;
}

in the scoped styles solves this issue i believe.

jesperlauridsen (approved):
I only have that little fix to support the aspect ratio - other than that this looks good!

Also, merge conflicts!

jorntx:
Merge conflicts fixed.

src/components/viewers/AudioVideo/VideoPlayer.vue
jesperlauridsen: Does it not work if it's in a new

.player {
aspect-ratio: 4/2;
}

and _NOT_ .video-player ?

jesperlauridsen: The idea comes from https://www.vidstack.io/docs/player/styling/introduction?styling=default-theme, where it seems you can style the player at will, because the web components they use aren't closed off.

jesperlauridsen: image
It looks right here, when i add a new class .player, and put the aspect ratio into that.

jorntx: You are so right - thanks a lot.

Added it and committed

jesperlauridsen: Looks good. Merging.

PR #98, opened 2023-11-28
Bump typescript to fix broken build due to vidstack

jorntx:
This took some time but found a fix for the build error caused by Vidstack. This has triggered a new warning about unsupported TypeScript version. Hopefully I can address this issue in the package maintenance PR.

QA --> check if you can build the project.


jesperlauridsen:
Seems weird. I guess we should always test if we can build if we're messing with the underlying stuff for the app. Good point. Tested that main couldn't build, and verified that I can with this fix. Approved.

PR #101, opened 2023-12-03
adding sort to search results

jesperlauridsen:
This should do the trick.

It took some tinkering to get this working - and what we can sort on is not really established yet. But this adds he outline for it.


src/components/search/Sort.vue
```diff
@@ -0,0 +1,136 @@
...
+			<span class="btn-text">
+				<span ref="currentSort">
+					{{
+						locale === 'da'
+							? searchResultStore.sort !== ''
+								? t('search.sortAfter') + searchResultStore.sort.split('%20')[0]
+								: t('search.sortBy')
+							: searchResultStore.sort !== ''
+							? t('search.sortAfter') + searchResultStore.sort.split('%20')[0]
+							: t('search.sortBy')
```
jorntx, line 16: Just balancing on a fine line where this code could be moved to a separate method :)

The template is not to large so you could opt for keeping it there this time.

src/views/Search.vue
```diff
@@ -215,6 +222,12 @@ export default defineComponent({
 				if (checkParamUpdate(newp, prevp) && route.query.q !== undefined) {
 					searchResultStore.setFiltersFromURL(route.query.fq as string[]);
 					searchResultStore.setStartFromURL(route.query.start as string);
+					// we wanna reset the sort if it's a NEW search started (which we determine is when you use a new query).
```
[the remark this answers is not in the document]
jesperlauridsen: Agreed. We'll leave it at this untill we know more. Personally, i like it this way - but it's not a hill I'm willing to die on.

src/components/search/Sort.vue
jesperlauridsen: I vote for keeping it this way for now - because _HOPEFULLY_ we can get rid of this way to look at locale when we dig deeper into the i18n plugin at some point :pray:

jesperlauridsen:
Okey. This took a little bit of brainpower to get sorted - sorry about the delay.

I've reworked it a little bit. the reset of the sort in Search.vue was not valid - since it didn't update the params of the URL (Which make sense, because the sort check was promted by a change in the URL.)

Instead I've reworked so the check is made before we actually change the URL. It works for now in the SearchBarWrapper, but if any more of this kinda stuff needs to be worked into the URL handling when searching, it should probably be made into a proper method instead of being done on the fly just before pushing the route.

jorntx (approved):
Looks good - will merge:)

PR #102, opened 2023-12-06
Change error message on 403

jorntx:
Small PR - user now gets 'not allowed' when encountering 403 in full record but only in the overlayed message 'toast'. We need a design for this use case for the whole full record and I already made a separate ticket for this:)

### DIFF
--- a/src/views/ShowRecord.vue
+++ b/src/views/ShowRecord.vue
@@ -59,9 +59,7 @@ export default defineComponent({
 			try {
 				return await APIService.getRecord(id);
 			} catch (err) {
-				const errorMsg =
-					(err as AxiosError).response?.status === 403 ? t('error.record.notAllowed') : t('error.record.loadingFailed');
-				errorManager.submitError(err as AxiosError, errorMsg);
+				handleShowRecordError(err as AxiosError, 'recordCall');
 			}
 		};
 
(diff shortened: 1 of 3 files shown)


src/views/ShowRecord.vue
```diff
@@ -59,7 +59,9 @@ export default defineComponent({
 			try {
 				return await APIService.getRecord(id);
 			} catch (err) {
-				errorManager.submitError(err as AxiosError, t('error.getrecordfailed'));
+				const errorMsg =
```
jesperlauridsen, line 62: I'm not opposed to this - but maybe this should be a seperate function getting the code, and then determining what error message should be thrown. That way, it will be easier to expand on it in the future, perhaps?

jesperlauridsen (approved):
I only have the one thought - all else is very nice and clean!

src/views/ShowRecord.vue
jorntx: I think you have a valid point and gave it some thought. We could abstract this out but are we sure that we always want to handle the 403 in the same way (spoiler: no)? If not then the decision of how it is to be handled is still made here directly or indirectly.
Even though this might be the case we could still abstract it out but that would require that we provide enough 'origin context' so the handler will know exactly what to do.
I would appreciate to have your thoughts on this.

jorntx: Come to think of it - maybe I just over-engineered because I misread your point - did you mean a separate method in ShowRecord.vue? Because that makes sense in any case:)

jesperlauridsen: My original thought was to keep it in ShowRecord, and just make a small function to handle this. I think the logic of how you wanna throw an error should be handled where it's going to happen - in this case in the ShowRecord. So imho, as simple function here to handle what kinda error you throw based on the code would suffice. And that could be expanded base on whatever we need later on in ShowRecord.

jorntx: Cool - totally valid - forget my first comment (and focus on the second). Will fix.

jorntx: Fixed it and much better - thanks!

jesperlauridsen (approved):
Nice rework. I like it!

PR #104, opened 2023-12-12
Disc 577/linting issues

jorntx:
As the title says. Fixed som small lint issues but even so I had to monkey around with the setup to get it to lint (removed prettier as extension while keeping it as plugin).

Give it a spin and see if you can lint/build.


jesperlauridsen:
Building and serving works just fine. I don't get anything weird, so I'm approving and merging this. good job!

PR #105, opened 2023-12-12
new values for the delimitation

jesperlauridsen:


jorntx (approved):
Very straight forward. Took it for a spin and works as intended.

PR #106, opened 2024-01-10
First stab at eslint fix and some linting too

jorntx:
First stab at fixing our crazy linter problems...

Tinkered with the eslint configuration and reintroduced the eslint-prettier integration.

It would be nice if you tried the following:

First clear node_modules and run `npm install`. You will be prompted if you try this with the wring node version.

Then:

- Mess with indention in some files and press save and see that it is fixed instantly to your liking
- Try to remove a semicolon from a line and pres save and see that it is fixed instantly to your liking
- Try to change some single quotes to double quotes press save and see that it is fixed instantly to your liking
- Try to remove a dangling (e.g. line 92 in SearchResults.vue) press save and see that it is fixed instantly to your liking

Finally try `npm run lint` and see that it runs without a hitch. If you go the final mile than try to mess up some indention and then run `npm run lint` and see that it gets auto fixed.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,11 +3,7 @@ module.exports = {
 	env: {
 		node: true,
 	},
-	extends: [
-		'plugin:vue/vue3-essential',
-		'eslint:recommended',
-		'@vue/typescript/recommended',
-	],
+	extends: ['plugin:vue/vue3-essential', 'eslint:recommended', '@vue/typescript/recommended'],
 	plugins: ['prettier'],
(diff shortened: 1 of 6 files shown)


jesperlauridsen:
This all seems to be working super awesome! I'm very happy with this! Good job! All my testing was passed with flying colors!
Im gonna merge this right away and update my branch with this because oh lord i've missed this for the last day!

jesperlauridsen (approved):

PR #107, opened 2024-01-14
Disc 526/new visual loading animation

jesperlauridsen:
Okey, so, this has been through a few reworks, and its still missing a little piece, with the last of the facets being loaded as a list of filters below the toggles, but I feel like the rest is in a state where I would like some second eyes on it.

I think it works pretty well now - even though our index is blazing fast so you don't get to see it much. A setTimeout can be put in to prolong the load so you can play with it a little more.

I ended up using gsap for this - the transitionend events are buggy, and dont seem to trigger in the same way every time. I don't know if my timings are just too close so the cycles somtimes mess up, but going with gsap fixed all of it.

So, the components wc-facet-item and wc-result-item both have 2 states now, a content and skeleton state, so to speak. And we use gsap to help toggle between those. The Vue lists for facets and results handle the in/out of the components now.

Would like any and all thoughts on this. :pray:


src/components/search/Facets.vue
```diff
@@ -42,34 +48,44 @@ export default defineComponent({
...
+									currentFacets.value = newFacets;
+									channelFacets.value = simplifyFacets(newFacets['creator_affiliation']);
+									currentFacetNr.value = channelFacets.value.length;
+									lastUpdate.value = new Date().getTime();
+									showFacets.value = true;
+								},
+								sum.length <= 0 ? 0 : 600,
+							);
+						}
+					}, 0);
```
jorntx, line 88: `setTimeout` inside `setTimeout` now you are cooking with butter!

src/components/search/SearchResults.vue
```diff
@@ -1,27 +1,53 @@
...
+							: ''
+					"
+					:starttime="
+						currentResults[index]
+							? currentResults[index].startTime !== undefined
+								? getBroadcastDate(currentResults[index].startTime as string, locale) +
+								  ' ' +
+								  t('record.timestamp') +
+								  getBroadcastTime(currentResults[index].startTime as string)
+								: t('record.noBoardcastData')
```
jorntx, line 44: Seems like we both missed a typo a couple of months ago

src/components/search/Facets.vue
jesperlauridsen: Hah, that's gonna be removed asap. It was for testing purposes only!

src/components/search/SearchResults.vue
jesperlauridsen: I had to read that 3 times to find out what you were talking about. Will fix!

src/components/search/wc-facet-checkbox.ts
```diff
@@ -41,13 +44,64 @@ class checkboxComponent extends HTMLElement {
...
+		const loading = this.shadow.querySelector('.loading') as HTMLDivElement;
+		const shimmer = this.shadow.querySelector('.shimmer') as HTMLDivElement;
+		shimmer.style.animationDelay = Math.random() * 1 + 's';
+		loading && (loading.style.display = 'flex');
+		gsap.to(loading, {
+			opacity: 1,
+			delay: 0.01,
+			duration: 0.25,
+		});
+	}
```
jorntx, line 80: There is a potential for condensing this by making it one method with a boolean 'switch' but I am torn here as this is very readable as it is.

src/components/search/wc-result-item.ts
jorntx, line 1: Good stuff, and the same comments for wc-result-item.ts are valid here too

src/components/search/wc-searchbar.ts
```diff
@@ -249,6 +260,7 @@ const SEARCH_COMPONENT_TEMPLATE = /*html*/ `
 								<label for="focusSearchInput" class="sr-only">Søg på KB.dk</label>
 								<input type="search" id="focusSearchInput" class="form-control" placeholder="Søg på KB.dk" name="simpleSearch">
 							</div>
+							<div class="spinner-container"><div class="spinner"></div></div>
```
jorntx, line 263: Nice visual touch and love you included the aria-busy attribute already

src/components/search/wc-searchbar.ts
```diff
@@ -369,7 +381,42 @@ const SEARCH_COMPONMENT_STYLES = /*css*/ `
...
+			[aria-busy] {
+			--spinner-animation-play-state: running;
+			}
+			@keyframes spin {
+				from {
+					transform:rotate(0deg);
+				}
+				to {
+					transform:rotate(360deg);
+				}
```
jorntx, line 418: holy smokes - seeing things here I'm having a hard time deciphering, had to look a couple of things up to grasp it:) Looks so nice when running, though :)

src/types/GenericRecordTypes.ts
```diff
@@ -68,3 +68,8 @@ export interface PublicationType {
 		broadcastDisplayName: string;
 	};
 }
+
+export interface facetPair {
```
jorntx, line 72: facetPair --> FacetPair :)

src/views/Search.vue
```diff
@@ -163,6 +163,8 @@ export default defineComponent({
 	},
 
 	setup() {
+		const mockdata1 = ['1', '2', '3', '4', '5', '6', '7', '8'] as unknown as GenericSearchResultType[];
+		const mockdata2 = ['1', '2', '3'] as unknown as GenericSearchResultType[];
```
jorntx, line 167: Yes, much better

src/utils/filter-utils.ts
```diff
@@ -0,0 +1,78 @@
...
+	if (title && key) {
+		const searchResultStore = useSearchResultStore();
+		const filterString = `${key}:"${title}"`;
+		return searchResultStore.filters.includes(`fq=${encodeURIComponent(filterString)}`);
+	} else {
+		return false;
+	}
+};
+
+const cloneRouteQuery = (route: RouteLocationNormalizedLoaded) => {
```
jorntx, line 19: One could argue that `cloneRouteQuery` never will feel at home here in` filter-utils.ts` :)

jorntx: If this is the only usage I'm open to keeping it here for now

src/utils/filter-utils.ts
jorntx, line 1: I really digg that you moved these filter methods to a separate file

jorntx (commented):
Very very good job! I really like the visuals of this and I think it ended up much cleaner than the first couple of passes you took at this. 

I left a few comments and found one bug when testing.

Try to search for 'Peter' and select facet 'Kalundborg' and then watch what happens to the 'Releterede emner' box.

src/components/search/Facets.vue
```diff
@@ -42,33 +54,50 @@ export default defineComponent({
...
+									currentFacets.value = newFacets;
+									channelFacets.value = simplifyFacets(newFacets['creator_affiliation']);
+									categoryFacets.value = simplifyFacets(newFacets['categories']);
+									currentFacetNr.value = Math.min(channelFacets.value.length, 10);
+									lastUpdate.value = new Date().getTime();
+									showFacets.value = true;
+								},
+								sum.length <= 0 ? 0 : 600,
+							);
+						}, 5000);
```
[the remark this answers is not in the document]
jesperlauridsen: Agreed. Will do!

src/components/search/Facets.vue
```diff
@@ -27,12 +32,19 @@
 import { defineComponent, PropType, onMounted, ref, onUnmounted, watch } from 'vue';
 import { useSearchResultStore } from '@/store/searchResultStore';
 import { FacetResultType } from '@/types/GenericSearchResultTypes';
-import { useRoute, useRouter, RouteLocationNormalizedLoaded } from 'vue-router';
+import { useRoute, useRouter } from 'vue-router';
+import CategoryTags from '@/components/search/CategoryTags.vue';
+import { createFilter, addFilter, removeFilter, filterExists, simplifyFacets } from '@/utils/filter-utils';
```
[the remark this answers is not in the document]
jesperlauridsen: It seemed like the best way to do this, since they share the functionality (and perhaps in the future other components will to)

src/components/search/wc-facet-checkbox.ts
```diff
@@ -41,13 +44,64 @@ class checkboxComponent extends HTMLElement {
...
+
+	private containerRevealed() {
+		const container = this.shadow.querySelector('.checkbox-container') as HTMLDivElement;
+		container && (container.style.display = 'flex');
+		gsap.to(container, {
+			opacity: 1,
+			delay: 0.01,
+			duration: 0.25,
+		});
+	}
```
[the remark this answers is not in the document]
jesperlauridsen: I did a little refactoring of it - I like having it in separate methods here.

src/components/search/wc-facet-checkbox.ts
jesperlauridsen: I did a little refactoring of it - I like having it in separate methods here.

jorntx (approved):
Thanks for the fixes - this runs smooth and nice implementation.

Also tested the build - just in case:)

 Hope the UX crew digg it!

PR #109, opened 2024-01-30
Disc 669 search hardening

jesperlauridsen:
Okey, so this grew a little (because it always does these days). The following is done:

The searchbar now locks for 500 ms when you search. Everything in there locks, except the reset button. I think that's appropriate, so you can always reset. Furthermore, facets, categories, sorting and the pager locks, so you cant use them. They all use the searchResultStore.loading variable as indicator for when to open/close.

Now, for this to work, i also had to rework the way the skeletons were show - they now use the same principle, with the loading variable.

I've added a small function for delay on the api calls, only for development tho. Should not be an issue to have it in there - but we can discuss this later too.

I also did some refactoring here and there aswell, removing some unused variables etc.

But, the really interesting thing is the in the function for getting the new results and facets. I've made it so the check is checking for the 4 variables that can change. And since we always have the origin:broadcaster filter, it's just poped to remove it. This changes a little bit how the structure of the variable is (since it, if there was only one filter, would be returned as a string, instead of an array from the solr), but I think this way actually covers that part neatly. The check *seems* solid, but I would really love your take on this.

Tested it fairly thoroughly, but I would appreciate you giving it a swing aswell at some point.

### DIFF
--- a/src/components/search/wc-searchbar.ts
+++ b/src/components/search/wc-searchbar.ts
@@ -109,14 +109,13 @@ class SearchBarComponent extends HTMLElement {
 			const group = this.shadow.querySelector('.rdl-advanced-search');
 			const inputs = group?.querySelectorAll('input') as unknown as Array<HTMLInputElement>;
 			const sbtn = this.shadow.querySelector('#searchButton');
-			console.log(group, inputs);
 			if (newValue === 'true') {
 				group?.classList.add('locked');
 				sbtn?.setAttribute('disabled', 'disabled');
 				inputs?.forEach((input) => {
 					input.setAttribute('disabled', 'disabled');
 				});
-				console.log('searchbar temporarily disabled.');
+				//console.log('searchbar temporarily disabled.');
 			} else {
 				group?.classList.remove('locked');
 				sbtn?.removeAttribute('disabled');
@@ -124,7 +123,7 @@ class SearchBarComponent extends HTMLElement {
 					input.removeAttribute('disabled');
 				});
 				inputs[0].focus();
-				console.log('searchbar enabled.');
+				//console.log('searchbar enabled.');
 			}
 		}
 		if (name === 'reset-value') {
@@ -139,7 +138,7 @@ class SearchBarComponent extends HTMLElement {
 		}
 		if (name === 'spinner') {
 			const spinner = this.shadow.querySelector('.spinner') as HTMLDivElement;
-			console.log(spinner);
+			//console.log(spinner);
 			if (newValue === 'true') {
 				spinner.setAttribute('aria-busy', 'true');
 				spinner.style.opacity = '1';
--- a/src/components/search/wc-searchbar.ts
+++ b/src/components/search/wc-searchbar.ts
@@ -74,7 +74,7 @@ class SearchBarComponent extends HTMLElement {
 	}
 
 	static get observedAttributes() {
-		return ['reset-value', 'q', 'lang', 'spinner'];
+		return ['reset-value', 'q', 'lang', 'spinner', 'disable-search'];
 	}
 
 	private getPresetFilter(key: string): string {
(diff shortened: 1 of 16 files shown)


jesperlauridsen:
So, now the comparison is made on an uuid instead - which is _A LOT_ nicer, imho!

src/components/search/wc-searchbar.ts
```diff
@@ -319,7 +340,21 @@ const SEARCH_COMPONMENT_STYLES = /*css*/ `
 		-webkit-font-feature-settings: 'liga';
 		-webkit-font-smoothing: antialiased;
 	  }
+		
+		#searchButton, #focusSearchInput, .rdl-advanced-radio {
+		transition: color 0.3s linear 0s;
+		}
+
+		.locked {
+				/* https://jxnblk.github.io/grays/ */
```
jorntx, line 349: Yes!

src/store/searchResultStore.ts
```diff
@@ -94,43 +96,62 @@ export const useSearchResultStore = defineStore('searchResults', () => {
...
 	const getSearchResults = async (query: string) => {
 		let searchFilters = '';
 		if (filters.value.length > 0) {
 			filters.value.forEach((filt: string) => {
 				searchFilters += `&${filt}`;
 			});
 		}
+		const startParam = start.value === '' ? '' : `&start=${start.value}`;
+		const sortParam = sort.value === '' ? '' : `&sort=${sort.value}`;
+
```
jorntx, line 112: Lets put a note here to future coder about why we use URL.createObjectURL

src/store/searchResultStore.ts
```diff
@@ -94,43 +96,62 @@ export const useSearchResultStore = defineStore('searchResults', () => {
 		filters.value.splice(filters.value.indexOf(filter), 1);
 	};
 
+	const validateCurrentSearch = (uuid: string): boolean => {
```
jorntx, line 99: Maybe a slightly better name here, so one knows what is validated. Maybe something along the lines of `responseMatchesCurrentSearch` or `isCurrentSearchStillValid`?

src/store/searchResultStore.ts
```diff
@@ -4,12 +4,14 @@ import { APIService } from '@/api/api-service';
 import { useSpinnerStore } from '@/store/spinnerStore';
 import { ErrorManagerType } from '@/types/ErrorManagerType';
 import { AxiosError } from 'axios';
-import { inject, ref } from 'vue';
+import { inject, ref, toRaw } from 'vue';
```
jorntx, line 7: `toRaw` seems to be unused

src/store/searchResultStore.ts
```diff
@@ -4,12 +4,14 @@ import { APIService } from '@/api/api-service';
...
 import { useI18n } from 'vue-i18n';
 import { FacetResultType } from '@/types/GenericSearchResultTypes';
 import { LocationQueryValue } from 'vue-router';
 
 export const useSearchResultStore = defineStore('searchResults', () => {
+	const currentSearchUUID = ref('');
+	const comparisonSearchUUID = ref('');
```
jorntx, line 14: These are not returned from the store - is there any reason not to use good ol' `let` 
```
let currentSearchUUID = ''; 
let comparisonSearchUUID = '';
```

src/types/APIResponseTypes.ts
```diff
@@ -30,18 +31,22 @@ export interface APIMoreLikeThisResponseType {
...
-				wt: string;
-				indent: string;
-				facet: string;
-			};
-		};
+		responseHeader: APIResposeHeaderType;
+	};
+}
+
+export interface APIResposeHeaderType {
```
jorntx, line 38: Good separation. Small typo:)

src/views/Search.vue
```diff
@@ -41,7 +41,13 @@
 								</span>
 							</button>
 						</div>
-						<div v-else-if="searchResultStore.searchFired">{{ $t('search.nohit') }}</div>
+						<div v-else-if="searchResultStore.searchFired">
+							<div
+								class="buffer"
+								v-if="searchResultStore.loading === true"
```
jorntx, line 47: Could be shortened to `v-if="searchResultStore.loading"`

src/views/Search.vue
```diff
@@ -228,9 +234,7 @@ export default defineComponent({
 					searchResultStore.setFiltersFromURL(route.query.fq as string[]);
 					searchResultStore.setStartFromURL(route.query.start as string);
 					searchResultStore.setSortFromURL(route.query.sort as string);
-					searchResultStore.getSearchResults(route.query.q as string).then(() => {
-						window.scrollTo({ top: 0, behavior: 'smooth' });
```
jorntx, line 232: It is actually settling for the eyes that this was moved:)

src/views/Search.vue
jesperlauridsen: Fixing it!

src/types/APIResponseTypes.ts
jesperlauridsen: Fixing it!

src/store/searchResultStore.ts
jesperlauridsen: Hm. Fixing it. Is it better to do than having them as const values?

src/store/searchResultStore.ts
jesperlauridsen: Fixing it!

src/store/searchResultStore.ts
jesperlauridsen: Better name, fixing it!

jesperlauridsen:
Added the fixes. Also a few bugfixes for the sort and pager.

jorntx (approved):
Thanks for the changes and bug fix. Looks good now. Took it for a spin and the debounce og handling of multiple requests works fine. Even threw a few filters in there:)

PR #110, opened 2024-01-30
Trimmed locale handling

jorntx:
Took a look at the locale handling and all in all the locale handling is implemented fine. So I:

- Trimmed some of the locale handling code in vue components
- Removed superfluous code

Going forward we need to be aware when we:

- find ourselves checking for locale in templates in Vue components (usually a ternary checking for 'da')
- find ourselves 'watching' locale in vue components - typically child components

### DIFF
--- a/src/components/common/Duration.vue
+++ b/src/components/common/Duration.vue
@@ -1,9 +1,11 @@
 <template>
-	<span>({{ formattedDuration }})</span>
+	<span>
+		<span>({{ formattedDuration }})</span>
+	</span>
 </template>
 
 <script lang="ts">
-import { defineComponent, PropType, ref, onMounted, watch } from 'vue';
+import { defineComponent, PropType, computed } from 'vue';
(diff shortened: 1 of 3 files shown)


jesperlauridsen:
This looks great. And works. Wonder why we ever needed the other approarch. In addition to this fine work I've found a few places where we don't translate yet. I'm mentioning them here, because JIRA is down. We should probably move them there once it's up again:

Searchpage:

Search-button in the input field.

recordpage:

Back button.
"Sendt" in the blue boks.

Additionally, the "Yderligere information om udsendelsen" in the accordian-boks is translate, but doesn't update reactively. It only reacts on reload. We should look at that also.

But the PR looks good. Main merged in, and we're good to go!

jesperlauridsen (approved):

jorntx:
Thanks for the review @jesperlauridsen   - main is merged in, so we a go for merge.

PR #111, opened 2024-01-30
DISC-359-show-if-on-dev-or-stage-env

jesperlauridsen:
The awesome title says it all!


jorntx (approved):
Apart from the dev. environment looks kinda 'dangerous' it works as intended :)

PR #112, opened 2024-01-30
DISC-760/Fix broken images

jesperlauridsen:
A fairly straight forward rework.

The image now takes one property - imagedata, which we observe. If it changed, we update all the stuff. For now, we don't really have any images to show, but the placeholder shows up, and everything reacts as it should. So this will work for now untill we get a better idea of how to handle images/thumbnails/players.

Also works for both the result-item and the spot-item on the frontpage, so that's good.

### DIFF
--- a/src/components/common/wc-image-item.ts
+++ b/src/components/common/wc-image-item.ts
@@ -1,12 +1,6 @@
 import { ImageComponentType } from '@/types/ImageComponentType';
 
-class ImageComponent extends HTMLElement implements ImageComponentType {
-	imgSrc: string | undefined;
-	imgTitle: string | undefined;
-	altText: string | undefined;
-	icon: string | undefined;
-	aspect: string | undefined;
-	placeholder: string | undefined;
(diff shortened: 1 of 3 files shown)


src/components/common/wc-image-item.ts
```diff
@@ -17,38 +11,40 @@ class ImageComponent extends HTMLElement implements ImageComponentType {
...
+				this.querySelector('.image-item')?.setAttribute('title', imageData.imgTitle);
+			}
+			if (imageData.icon) {
+				const iconHolder = this.querySelector('.type-symbol');
+				iconHolder && (iconHolder.textContent = imageData.icon);
+			}
+			if (imageData.aspect) {
+				const imageWrapper = this.querySelector('.image-wrapper') as HTMLElement;
+				imageWrapper && (imageWrapper.style.aspectRatio = imageData.aspect);
+			}
```
jorntx, line 47: Nice and flexible

jorntx (approved):
Good rework and lets roll with this for now

PR #113, opened 2024-02-05
Disc x/autocomplete in search bar

jesperlauridsen:
So, suggestions in the searchbar.

3 ways to interact with it. When you write more than 2 chars, it gets the suggestions. DR is a good test, to start with. Right now its case sensitive.

You can navigate it with the mouse, up and down keys if you have focus in the search input field, and tab, since they're made as buttons, so you can tab through them aswell. Accessability wise, there is maybe a little more work to do, but i think this is a very good starting point.

Give i a whirl and tell me what you think. I've tried to make sure it handles nice in every case. When you search it is removed, and when you search it _should_ stop the process of firing another check for suggestions after the lats keystroke.

### DIFF
--- a/src/components/search/SearchBarWrapper.vue
+++ b/src/components/search/SearchBarWrapper.vue
@@ -5,6 +5,7 @@
 		:background-img-url="backgroundImage"
 		:lang="locale"
 		:spinner="searchResultStore.loading"
+		:suggest="JSON.stringify(searchResultStore.AutocompleteResult)"
 	></kb-searchbar>
 </template>
 
@@ -23,9 +24,11 @@ export default defineComponent({
 		const searchQuery = ref('');
 		const preliminaryFilter = ref('');
 		const { t, locale } = useI18n();
+		const stopAutcomplete = ref(true);
 		const errorManager = inject('errorManager') as ErrorManagerType;
 		const router = useRouter();
 		const route = useRoute();
+		let AutocompleteTimer: ReturnType<typeof setTimeout> | null = null;
 
 		const searchResultStore = useSearchResultStore();
 		const xReset = ref(false);
@@ -37,6 +40,10 @@ export default defineComponent({
 			if (route.query) {
 				searchQuery.value = route.query.q as string;
 			}
+
+			window.onpopstate = function () {
+				searchQuery.value = route.query.q as string;
+			};
 		});
 
 		watch(searchQuery, (newValue) => {
@@ -84,6 +91,12 @@ export default defineComponent({
 			router.push({ name: 'Home' });
 		};
 
+		const getAutocompleteResponse = (query: string) => {
+			if (query !== undefined && query.length >= 2) {
+				searchResultStore.getAutocompleteResults(query);
+			}
+		};
+
 		const updateQuery = (e: CustomEvent) => {
 			e.stopPropagation();
 			e.preventDefault();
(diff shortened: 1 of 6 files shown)


src/store/searchResultStore.ts
```diff
@@ -92,15 +94,26 @@ export const useSearchResultStore = defineStore('searchResults', () => {
...
+		AutocompleteResult.value = [];
+	};
+
 	const removeFilter = (filter: string) => {
 		filters.value.splice(filters.value.indexOf(filter), 1);
 	};
 
+	const getAutocompleteResults = async (query: string) => {
+		const autocompleteReponse = await APIService.getAutocomplete(query);
+		const object = autocompleteReponse.data.suggest.dr_title_suggest;
```
jorntx, line 107: Lets find another name for this variabel:)

src/components/search/SearchBarWrapper.vue
```diff
@@ -131,6 +148,30 @@ export default defineComponent({
 			}
 		};
 
+		watch(
+			() => searchQuery.value,
+			(newStart: string, prevStart: string) => {
+				if (newStart !== prevStart) {
+					if (AutocompleteTimer !== null) {
```
jorntx, line 155: I don't think you need the` null` check here. `clearTimeout` will take care of that for you.

src/components/search/SearchBarWrapper.vue
```diff
@@ -131,6 +148,30 @@ export default defineComponent({
...
+						clearTimeout(AutocompleteTimer);
+					}
+					if (!stopAutcomplete.value) {
+						if (newStart?.length < 2) {
```
jorntx, line 159: Do you need the null/undefined option here?

jorntx (commented):
Looks good as a first stab and works fine. I hope the UX'ers can add some mojo to it - it is very "flat" and blends to much in with the search box imho:)

The wiring looks solid though and only have some minor suggestions for the code.

src/components/search/SearchBarWrapper.vue
```diff
@@ -24,9 +25,11 @@ export default defineComponent({
 		const searchQuery = ref('');
 		const preliminaryFilter = ref('');
 		const { t, locale } = useI18n();
+		const stopAutcomplete = ref(true);
 		const errorManager = inject('errorManager') as ErrorManagerType;
 		const router = useRouter();
 		const route = useRoute();
+		let AutocompleteTimer: ReturnType<typeof setTimeout> | null = null;
```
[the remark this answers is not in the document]
jesperlauridsen: Very much so. I had to spent quite some time to get this right!

src/components/search/wc-searchbar.ts
```diff
@@ -105,6 +155,35 @@ class SearchBarComponent extends HTMLElement {
...
+							this.dispatchUpdate(searchQueryInputField.value);
+							this.dispatchSearch();
+						});
+					}
+					list?.append(ac);
+				});
+			} else {
+				list && (list.innerHTML = '');
+			}
+		}
```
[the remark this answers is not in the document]
jesperlauridsen: Not much to do here really, in a web-component. Would be way nicer in a vue component...

src/components/search/wc-searchbar.ts
```diff
@@ -33,6 +35,23 @@ class SearchBarComponent extends HTMLElement {
...
+					this.moveSelectorUp();
+					break;
+				case 'Enter':
+					this.executeOnSelector(e);
+					break;
+				default:
+					break;
+			}
+		});
+
```
jorntx, line 54: Looks good - cleaned up nicely

jorntx (approved):
Thanks for the changes - looks good and still works.

The suggester results are no longer valid but that must be the index that is out of sync

PR #114, opened 2024-02-05
Added spinner for fullpost

jesperlauridsen:
And removed the delay for morelikethis and fullpost api calls


jorntx (approved):
Implementation looks and straight forward

PR #116, opened 2024-02-08
Lighthouse fixes

jesperlauridsen:
Fairly straight forward. Not anything really important here as of right now - just minor tweaks here and there.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -20,7 +20,11 @@
 			class="wipe"
 			ref="wipe"
 		>
-			<img :src="getImgServerSrcURL()" />
+			<img
+				title="Royal Danish Library"
+				alt="Logo of the Royal Danish Library"
+				:src="getImgServerSrcURL()"
+			/>
 		</div>
 	</div>
 	<div class="content">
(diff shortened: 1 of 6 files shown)


src/App.vue
```diff
@@ -20,7 +20,11 @@
 			class="wipe"
 			ref="wipe"
 		>
-			<img :src="getImgServerSrcURL()" />
+			<img
+				title="Royal Danish Library"
+				alt="Logo of the Royal Danish Library"
+				:src="getImgServerSrcURL()"
```
jorntx, line 26: Øjjj, we shouldn't miss this basic stuff in the first place...and I see that this is not the only place:)

src/components/search/CategoryTags.vue
```diff
@@ -177,7 +177,7 @@ h2 {
 .category-tags .tag .tag-number {
 	font-size: 10px;
 	margin-left: 5px;
-	color: rgba(130, 130, 130, 1);
+	color: #383838;
```
jorntx, line 180: Nice to know

jorntx (approved):
Thanks for the changes and taking the first stab at this.

It wasn't awful but still nice to catch these things.

PR #117, opened 2024-02-09
Disc 747/spell check

jorntx:
First Stab at spell check/did you mean function. I think we will expand this as the data gets more stable. I haven't utilised all the potentiel in the spell check data.

A few notes:
- Ignorer the out-commented part of the unit test. I couldn't get it to run due to the translation part need to be mocked and I didn't want to spend more time on it for now. The most important tests runs (check that we handle if prop is null).
- The 'lorem ipsum' in the 0-hits page is to comply with the design from UX - they want a placeholder there as there will be more text to guide the user when they get around to making a final 0-hits page design.

### DIFF
--- a/src/api/api-client.ts
+++ b/src/api/api-client.ts
@@ -64,7 +64,7 @@ export class APIServiceClient {
 		return await this.httpClient.get(
 			`search/?q=${encodeURIComponent(
 				query,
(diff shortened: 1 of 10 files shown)


jesperlauridsen (approved):
Great work! No notes!

PR #120, opened 2024-02-14
Disc 791/rewrite result item to vue

jorntx:
Complete rewrite wi-result-item from web component to Vue component.

Had to adjust wc-image-item along the way due to the need for open shadow DOM in order for it to work directly in the new Vue component (ResultItem).

Noticed during refactor and not fixed:
- The skeleton is for some reason not identical to the old one
- The skeleton animation is no longer synchronised with the facet animation
- There is no snazy gsap animation when transitioning from loading to content and vice versa. It fades in/out with Vue transition


src/components/search/wc-result-item.ts
```diff
@@ -1,399 +0,0 @@
...
-				imageData.imgSrc = this.data.thumbnail;
-				imageData.altText = this.data.title;
-				imageData.imgTitle = this.data.title;
-				imageData.placeholder = this.placeholder;
-				imageComponent.setAttribute('imagedata', JSON.stringify(imageData));
-			}
-		}
-	}
-
-	private updateSkeletonDifferences() {
```
jesperlauridsen, line 110: This part seems to have been omitted from the refactor. This made the skeletons have different lengths, so they looked "different". Both the spans and the titles. This probably needs to be rebuild with Vue tech instead of this way to do it.

jesperlauridsen (changes requested):
All looks good, apart from the missing skeleton sizes. Update that and it's good to go! \o/

src/components/search/wc-result-item.ts
jorntx: I gave it a stab. Check it out and see if it is OK.

jesperlauridsen:
This is super close to what is was before we remade the component. There was a little padding between the skeleton bars, but I think we can just ignore this for now and fix that when doing animations. Good job from my end - just some merge conflicts and we're good to go!

jesperlauridsen (approved):

jorntx:
Ready for merge @jesperlauridsen !

PR #121, opened 2024-02-15
Disc 825/maintenance 20240214

jorntx:
Small PR but tinkered with some critical parts here:

- Updated Vue version to 3.4.19 which has **BREAKING changes**. The app seems to have survived and I basw this on the fact that the automated tests still run and I gave the app a good QA and found nothing. Please give the app a good spin to check if everything is OK - especially with regards to transitions (the vue core code changed here - hopefully to the better).
- Updated VueI18n. No problems detected but try to switch lang a few times anyway:)
- Updated some small dev. post-css package (sec. audit)  - nothing to see as it is only patch-level. Just ignore the package-lock.json diff.
- Cleaned up dev and prod dependencies so they are correct and consistent
- Removed Openseadragon from the bundle as it is not used for now.
- Had to manually set the __VUE_PROD_HYDRATION_MISMATCH_DETAILS__ flag to silence the console warning. This should be set by default in vue-cli but no such luck yet.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,6 @@
         "axios": "^1.6.2",
         "gsap": "^3.12.2",
         "hls.js": "^1.4.12",
-        "openseadragon": "^3.1.0",
         "pinia": "^2.1.7",
         "vidstack": "1.6.1",
         "vue": "3.4.19",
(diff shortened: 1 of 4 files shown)


jesperlauridsen (approved):
This looks very nice. I took the app for a good spin - didn't find any issues anywhere (apart from a few dead backend services, but that's not our fault! ;))+

PR #123, opened 2024-02-20
Disc 792/refactor searchbar

jesperlauridsen:
So this became pretty big - the searchfield takes care of a lot of stuff, and some of it had to be moved around. I think it landed quite nicely - but I might have missed something. Feel free to give it a good whirl!


src/components/search/wc-searchbar.ts
jorntx, line 1: Remember to delete this file

src/views/Search.vue
```diff
@@ -90,6 +90,34 @@ export default defineComponent({
...
+						} else {
+							// General search error happened here so message to user should be generel
+							window.dispatchEvent(
+								new CustomEvent('component-error', {
+									detail: { customError: true, message: 'searchfailed', systemError: null },
+								}),
+							);
+						}
+					}
+				}
```
jorntx, line 120: There is nothing to grab the custom events here. I have made a PR with a suggestion for a rewrite to use the built in error manager. Merge it in if you like it. PR --> https://github.com/kb-dk/ds-web/pull/125

jorntx (approved):
Once again - so nice refactor - and thanks for the changes!

We are good to go here:)

PR #124, opened 2024-02-21
Simplify the route logic in search function

jorntx:
A suggestion for simplifying the route logic in the search function


jesperlauridsen (approved):
It's annoying how much better this is. Well done. *hat-tip*

PR #125, opened 2024-02-21
Refactor error handling code to use built in error manager

jorntx:
A suggestion for fixing the non working logic connected to error handling here using the built in error manager.


jesperlauridsen (approved):
I.. don't know how i didn't find this. This makes no sense. Did we never have a handler for this case? Anyway, good rewrite, no notes. 10/10.

PR #129, opened 2024-03-08
Migrate from vue-cli to Vite

jorntx:
This took some effort to switch vue-cli to Vite (below is the condensed change log) but I think it is worth it🎉 

What I want you to do is:

- Start with running` npm install` so you get all the new packages and get rid of vue-cli
- Now start the dev server NOTE - this is now `npm run dev` (starts on port 3000 is it is available) and test that the app is running as smooth as usual in every corner :) 
- Try to run npm run build and sheck that the app builds without hickups
- Try to run the deploy to devel script and see if that works as intended
- While doing all of the above work - don't fail to notice how BLAZINGLY FAST Vite is compared to Webpack 😎

**Please ignore all fails** in regards to lighthouse test and unit tests. This will be fixed in a later PR.

**Changes:**
Remove vue-cli dependencies and scripts
Add Vite build and dev scripts and dependencies
Tune typescript linting
Rewrite vue.config.js to vite.config.mts
Wipe webpack magic comments
Relocate contents of public folder
Rewrite all require statements (Vite only supports ESM)
Handle bundle import in index.html (no injection in Vite) 
Remove Babel config (not used)
Make environment variable imports conform to Vite


jesperlauridsen:
Everything looks very nice! Only thing i noticed was that the environment set in the top of the devel instance is missing. So that should be fixed :-)

index.html
```diff
@@ -4,17 +4,17 @@
     <meta charset="utf-8" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width,initial-scale=1.0" />
-    <link rel="icon" href="<%= BASE_URL %>favicon.ico" />
-    <title><%= htmlWebpackPlugin.options.title %></title>
+    <link rel="icon" href="/favicon.ico" />
+    <title>DR Arkiv</title>
```
jesperlauridsen, line 8: Hardcoded title now. I guess we're okey with this, no?

src/components/global/notification/NotificationItem.vue
```diff
@@ -29,7 +29,6 @@
 import { defineComponent } from 'vue';
 import { gsap } from 'gsap';
 import { useI18n } from 'vue-i18n';
-import GSAPTween from 'gsap';
```
jesperlauridsen, line 32: Nice find! ;)

index.html
jorntx: We are OK with it being hardcoded here, but we should set the title dynamically when the app is bootstrapped.

We haven't talked about what our strategy for titles is. I will make a placeholder issue for this.

jorntx:
> Everything looks very nice! Only thing i noticed was that the environment set in the top of the devel instance is missing. So that should be fixed :-)

Yes, good catch - I have fixed it now. If you hadn't spottet this, it would have made its way all the way to production. Maybe we should consider reversing the `isProduction` logic just in case :)

jesperlauridsen:
> > Everything looks very nice! Only thing i noticed was that the environment set in the top of the devel instance is missing. So that should be fixed :-)
> 
> Yes, good catch - I have fixed it now. If you hadn't spottet this, it would have made its way all the way to production. Maybe we should consider reversing the `isProduction` logic just in case :)

Yes, Good idea. You wanna do a JIRA on it, or do it here? Otherwise i'll accept this and merge to main. Would love to get this into main \o/

PR #130, opened 2024-03-12
DRA-162 add footer

jesperlauridsen:
This should add the footer. Note, that the cooke settings button doesn't work, we have a seperate JIRA for that.


src/components/global/nav/Footer.vue
```diff
@@ -0,0 +1,259 @@
...
+	[column: string]: {
+		title: string;
+		links: {
+			[link: string]: {
+				title: string;
+				link: string;
+			};
+		};
+	};
+}
```
jorntx, line 80: Why type locally in the footer component  and not with the rest of the types?

src/components/global/nav/wc-header-menu.ts
```diff
@@ -247,6 +247,7 @@ const MENU_COMPONMENT_STYLES = /*css*/ `
 	}
 
 	.rdl-logo {
+		/* TODO: We need a better way to get images into web components. */
```
jorntx, line 250: Agreed - can you make a ticket for it?

src/components/global/nav/Footer.vue
```diff
@@ -0,0 +1,259 @@
...
+
+export default defineComponent({
+	name: 'Footer',
+	setup() {
+		const { t, messages, locale } = useI18n();
+		const currentLocaleMessages = ref(undefined as unknown as Footer);
+
+		onMounted(() => {
+			currentLocaleMessages.value = toRaw(messages.value[locale.value].footer) as Footer;
```
jorntx, line 89: So this is how it ended. Eyebrow raised but eyes not bleeding:)

jorntx (commented):
Implementation looks good even with the "creative translation solution". 

I only have a few comments and a missing translation - see screenie :)

Screenshot 2024-03-14 at 07 55 16

src/components/global/nav/wc-header-menu.ts
jesperlauridsen: Will do!

jorntx (approved):
Thanks for the changes - this is good to go now.

PR #131, opened 2024-03-13
DRA-146 only loading player when the streaming url is there

jesperlauridsen:
So, I've added a check to see if we have a streaming URL. If we do, then we show the player. Otherwise we show a very neutral message of "we dont have the url for this".

This is just the first stab - i guess we need Design/UX to make a proper layout for this - but functionality-wise, this should be the trick.

Added for both radio and TV. Feel free to message me for 2 links, one that has an url, and one that doesn't :-) (I don't wanna share the uuids here for one of each type).


jorntx (approved):
Looks good and works beautifully, tested with both audio and video.

PR #132, opened 2024-03-14
Dra 164 cookie plugin

jesperlauridsen:
This should add the cookie plugin for now. it's a test instance. and since we're not really putting any cookies on, we should test this further when we get the script from kaltura running.

But as of now, it loads and shows up on init, and it can be found again in the footer, both on english and danish.


jorntx (approved):
Very straightforward - works fine and we can fine tune contents later as we discussed.

PR #133, opened 2024-03-20
A proposed fix for the footer transition bug

jesperlauridsen:
So, it worked really weird with the transitions when the footer was part of the main content, not being transitioned. So i propose that we build the footer into the different views. This makes the footer also transition in and out, which works really well. Take it for a spin, and look at the few proposed changes, and see if you're okey with this.


jorntx (approved):
That did the trick - transition now works:)

PR #135, opened 2024-03-22
Dra 60/kaltura player first draft

jorntx:
Adds support for Kaltura player when playing video (DRA-298)

This is a POC for video streaming, so the main task for now is to check the outline of the code and that it can stream


jorntx:
Closing this PR - as it turns out that I have bootstrapped the wrong player. Fresh PR will follow later.

PR #136, opened 2024-03-22
Should fix this.

jesperlauridsen:
Kinda nothing really to see here. It works on my machine now - hopefully it works everywhere \o/


jorntx (approved):
Yup, works for me too:)

PR #137, opened 2024-03-22
Dra 229/titles in browser tab

jesperlauridsen:


src/components/search/SearchBar.vue
```diff
@@ -234,7 +234,9 @@ export default defineComponent({
 
 		const setPreliminaryFilter = (value: string) => {
 			preliminaryFilter.value = value;
-			search();
+			if (searchResultStore.currentQuery !== '' && searchResultStore.currentQuery !== undefined) {
+				search();
+			}
```
jorntx, line 239: I gather you ran into a side quest that prompted this change:)

jorntx (commented):
This gets us going in the right direction and the code looks fine. 

I made a suggestion for making it a bit less locked in to DR archive. 

I still need to actually run it but as you know we are having issues right now so that will have to wait:)

jorntx:
Tested and works

PR #138, opened 2024-03-27
DRA-292 - Refactored to composition API.

jesperlauridsen:
Not much to if, if I understood this right o_O


jorntx:
Tested and refactor works as before

PR #140, opened 2024-03-27
DRA-90 - Should fix the opt-out

jesperlauridsen:
So, this should fix it, after reading through the docs here:
https://github.com/GoogleChrome/lighthouse/blob/main/docs/error-reporting.md

Delete your ~/.config/configstore/lighthouse.json to test if it indeed opts out. Another problem is that the .env file is no longer valid, since it states that Lighthouse should look at port 8080, instead of 3000 where it's now running. A fix has been added in that repo too.


jorntx (approved):
Looks good and works.

I fixed the config for lighthouse so it now runs on 3000 - just fetch a fresh copy from Aegis.

PR #141, opened 2024-03-27
DRA-230 - better check for the production logic

jesperlauridsen:
Also a very small PR. I also toned down the banner, because holy smokes. I really overdid that one :D


jorntx (approved):
So much better and safer and I see you tinkered a bit with the visuals too:)

PR #142, opened 2024-04-02
DRA OPTIONAL: 2 small fixes

jesperlauridsen:
A small fix to make the facets act right during second loading and beyond.

A small cookie script fix because that stuff keeps bugging me.


jorntx (approved):
I like your attention to detail:)

Good fix with the cookie script - bugged me as well that the bootstrapping of the app could go south with the script!

PR #143, opened 2024-04-04
Dra 60/kaltura player first draft

jorntx:
Finally a very basic POC on the video part. Give it go and see it works.

You will need a fresh conf from Aegis.


src/components/viewers/AudioVideo/VideoPlayer.vue
```diff
@@ -1,61 +1,103 @@
...
-		>
-			<media-provider></media-provider>
-			<media-video-layout></media-video-layout>
-		</media-player>
+			style="width: 1228px; height: 614px"
+		></div>
 	</div>
-	<!-- 	<div class="mobile-edge edge bottom"></div>
- -->
+	<div class="mobile-edge edge bottom"></div>
```
jesperlauridsen, line 10: Lucky for you there is a PR removing these ;)

src/components/viewers/AudioVideo/VideoPlayer.vue
```diff
@@ -1,61 +1,103 @@
 <template>
-	<!-- 	<div class="mobile-edge edge top"></div>
- -->
-	<div
-		v-if="src"
-		class="video-player"
-	>
-		<media-player
-			ref="player"
+	<div class="mobile-edge edge top"></div>
```
jesperlauridsen, line 2: Lucky for you there is a PR removing these ;)

src/components/viewers/AudioVideo/VideoPlayer.vue
```diff
@@ -1,61 +1,103 @@
...
-		const player = ref<MediaPlayerElement>();
-		const src = ref('');
-		onMounted(() => {
-			src.value = props.videoUrl ? props.videoUrl : '';
-			const cleanup = () => {
-				if (player.value) {
-					player.value.destroy();
+		const { t } = useI18n();
+		const errorManager = inject('errorManager') as ErrorManagerType;
+		let kalturaPlayer: PlayerType;
```
jesperlauridsen, line 38: you have a let kalturaPlayer here, and a const KalturaPlayer in the global scope before export default. Can we do something about that naming? I don't like that's its with a big and small 'k'. Unless you got a *REALLY* good explanation as to why this has to be this way?

jorntx: Good point - let me fix this.

src/components/viewers/AudioVideo/VideoPlayer.vue
```diff
@@ -1,61 +1,103 @@
...
+
+		const handleErrorDispatch = (type: string) => {
+			switch (type) {
+				case 'loadMedia': {
```
jesperlauridsen, line 40: This is very nice imho. However, right now I when these errors are thrown, they either get to be a persistent message (that needs to be closed), or a fleeting one, with at timer that dissapears. I think that should be a param sent with the error submission to the errorManager. Maybe a new JIRA for this?

jesperlauridsen (approved):
I think this looks very solid now. Good job. I think we need a JIRA for the error handler, so we can set when we submit an error, if it's sticky or fleeting. But it also needs to go around Daniel for some design overhaul i think

PR #146, opened 2024-04-10
Add application version to html source and linting

jorntx:
Not much to it. After a few roundtrips I settled on a very simple solution.

I also tweaked a few eslint and prettier settings to remove warnings/errors in index.html that where bugging me.

You need a new conf from Aegis to test it out


jesperlauridsen:
Solid. I got nothing here. Accepted and merged \m/

jesperlauridsen (approved):

PR #150, opened 2024-04-16
First go at implementing secure cookie flow

jorntx:
The very first POC of secure cookie flow. For now this only applies to thumbnail service. So the test is that thumbnails still work on this branch. Remember a fresh conf from Aegis.

Change log: 
Add request for auth cookie at app init
Switch to new proxy endpoint that now handles all thumbnail and authentication requests


jesperlauridsen (approved):
Took it for a spin. Looks good. And also very nice codewise, imho

PR #152, opened 2024-04-18
DRA-477 - small change to get icons on the main thumbnails

jesperlauridsen:


jorntx (approved):
Looks good. Only one small hiccup - I will chat you

PR #153, opened 2024-04-18
Dra 293/security fix for vhtml

jesperlauridsen:
Not much to it, but it should do the trick we won't have to worry about nasty stuff in the autocomplete


jorntx (approved):
Looks good - added comment and eslint rule disable for v-html.

PR #154, opened 2024-04-18
First stab at removing DR specifics

jorntx:
Stright forward as title says. Remember new aegis conf

### DIFF
--- a/src/api/api-client.ts
+++ b/src/api/api-client.ts
@@ -54,13 +54,13 @@ export class APIServiceClient {
 
 	//Search and record methods
 	async getFacetResults(query: string, filters: string, start: string, sort: string): Promise<APISearchResponseType> {
-		//Temporary fix/implementation for limiting to DR material
-		const DRLimiter = encodeURIComponent('broadcaster:"DR"');
+		//Temporary fix/implementation for limiting material
+		const broadcastLimiter = encodeURIComponent(import.meta.env.VITE_BROADCASTER_LIMITER);
 
 		return await this.httpClient.get(
 			`search/?q=${encodeURIComponent(
 				query,
-			)}&facet=true${filters}${start}${sort}&fq=${DRLimiter}&rows=0&facet.limit=25`,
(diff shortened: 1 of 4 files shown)


src/api/api-client.ts
```diff
@@ -71,11 +71,13 @@ export class APIServiceClient {
...
-		//Temporary fix/implementation for limiting to DR material
-		const DRLimiter = encodeURIComponent('broadcaster:"DR"');
+		//Temporary fix/implementation for limiting material
+		const broadcastLimiter = encodeURIComponent(import.meta.env.VITE_BROADCASTER_LIMITER);
 
 		return await this.httpClient.get(
-			`search/?q=${encodeURIComponent(query)}&facet=false${filters}${start}${sort}&queryUUID=${uuid}&fq=${DRLimiter}`,
+			`search/?q=${encodeURIComponent(
+				query,
+			)}&facet=false${filters}${start}${sort}&queryUUID=${uuid}&fq=${broadcastLimiter}`,
```
jesperlauridsen, line 80: Since this is moved to a config now, someone (or us in an alternate universe) might run this with no broadcastLimiter. If so, is it okey for the solr to recieve a fq= ... nothing? Or should the whole fq= be conditional to the existence of a broadcastLimiter variable that is filled? I'm just asking questions here!

jorntx: You Sir, are asking very good questions here. I will fix. This should of course be more defensive.

src/components/common/PortalContent.vue
```diff
@@ -26,7 +26,7 @@
 			<div class="edge blue"></div>
 
 			<div class="container">
-				<h3>Om DR Arkivet</h3>
+				<h3>Om arkivet</h3>
```
jesperlauridsen, line 29: This seems like placeholder stuff, so it'll probably be removed at some point in the future when we get real data. So this is fine for now i think

src/api/api-client.ts
jorntx: It is fixed now

PR #155, opened 2024-04-18
DRA-461 show pages in top of search results

jesperlauridsen:
Not much to it.

### DIFF
--- a/src/components/search/SearchOverhead.vue
+++ b/src/components/search/SearchOverhead.vue
@@ -12,16 +12,19 @@
 					</div>
 					<Sort v-if="searchResultStore.searchResult.length > 0" />
 				</div>
-				<button
-					v-if="searchResultStore.searchResult.length > 0"
-					class="filter-button"
-					@click="searchResultStore.toggleShowFacets(!searchResultStore.showFacets)"
-				>
-					<span class="material-icons">tune</span>
-					<span class="filter-button-text">
-						{{ showFacets ? $t('search.hideFilters') : $t('search.showFilters') }}
-					</span>
-				</button>
+				<div class="filter-options">
+					<button
+						v-if="searchResultStore.searchResult.length > 0"
+						class="filter-button"
+						@click="searchResultStore.toggleShowFacets(!searchResultStore.showFacets)"
+					>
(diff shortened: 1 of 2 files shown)


src/components/search/SearchOverhead.vue
```diff
@@ -12,16 +12,21 @@
...
+						@click="searchResultStore.toggleShowFacets(!searchResultStore.showFacets)"
+					>
+						<span class="material-icons">tune</span>
+						<span class="filter-button-text">
+							{{ showFacets ? $t('search.hideFilters') : $t('search.showFilters') }}
+						</span>
+					</button>
+					<div class="page-count">
+						side {{ (Number(searchResultStore.start) + 10) / 10 }}/{{ Math.ceil(searchResultStore.numFound / 10) }}
+					</div>
```
jorntx, line 28: Good simple implementation that works calculation wise. If the logic gets slightly more complicated you could consider moving it to functions in setup()

src/components/search/SearchOverhead.vue
```diff
@@ -74,6 +86,11 @@ export default defineComponent({
 	margin-left: 0;
 }
 
+.page-count {
+	font-size: 16px;
+	//color: #000000;
```
jorntx, line 91: You need to use asterisk `/*`  when commenting out in css:)

jorntx (changes requested):
Nice simple implementation. A couple of minor comments and one bug:

When loading the count shows 1/0 on page one and 2/0 on page two etc. - see attached screenie.
Screenshot 2024-04-19 at 10 22 22

Can be reproduced by searching for * and if you have search result you can step to page 2 and try to reload.

jesperlauridsen:
> Nice simple implementation. A couple of minor comments and one bug:
> 
> When loading the count shows 1/0 on page one and 2/0 on page two etc. - see attached screenie. Screenshot 2024-04-19 at 10 22 22
> 
> Can be reproduced by searching for * and if you have search result you can step to page 2 and try to reload.

That's a booboo! I'm sorry, fixed now!

jorntx (approved):
Thanks for the changes. Looks good and works as intended.

PR #156, opened 2024-04-19
DRA-380: title logic expanded

jesperlauridsen:
This should take care of the stuff we talked about. Let me know if you find any issues (or dont like the structure!)


jorntx (approved):
Nice rework. Works fine and now the ball is in the UX court:)

PR #159, opened 2024-04-25
DRA-173: zero hit design implementation

jesperlauridsen:
A few notes here:

Not all of the design is implemented here, since this is on a wireframe level. However, I've tried to take a constructive approach here, making components that are fully reusable. In this, I've build the following components:

A divider that can tilt both right and left, and have a headline
A component that has image and text side by side
A fullcover element with tilted edges that can hold anything

I've tried using them a little but, but stopped putting more in since it's all dummy data so far. But i think they show what they can be used for right now.

Just do a random search that generates zero hits, and you could be able to view it all.


src/components/global/content-elements/TiltedDivider.vue
```diff
@@ -0,0 +1,77 @@
...
+			},
+		},
+	},
+
+	setup(props) {
+		const dividerRef = ref<HTMLDivElement | null>(null);
+
+		onMounted(() => {
+			if (dividerRef.value) {
+				console.log(props);
```
jorntx, line 36: :)

src/components/global/content-elements/TextAndImage.vue
```diff
@@ -0,0 +1,44 @@
...
+			type: Boolean as PropType<boolean>,
+			default() {
+				return true;
+			},
+		},
+	},
+
+	setup() {
+		return {};
+	},
```
jorntx, line 24: Just delete this as it is not needed

src/components/global/content-elements/EdgedContentArea.vue
jorntx, line 1: Nice clean component with good conf options (props).

jorntx (approved):
Nice clean (POC) implementation of zero hits. I have nothing to add except some very basic clean up.

jorntx (approved):

PR #160, opened 2024-04-26
DRA-566/ bug in preliminary filters

jesperlauridsen:
This should remove the bug with the preliminary filters not being placed correctly when re-searching after choosing one of them.


jorntx (approved):
Looks good and fixed the bug

PR #161, opened 2024-05-02
first stab at making some generic thumbnails for audio posts.

jesperlauridsen:
This is a draft until we get some feedback from design/UX. This has just been put in as a draft so we don't forget it.


jorntx (commented):
Nice implementation!

Can we do something about the variation of the bars - they look a bit similar to me?

jorntx (approved):
Looks good and I noticed the more diverse images:) 

Had one minor comment that needs no action. When you hit the thumbs up (or down) on that I will merge.

PR #163, opened 2024-05-14
Dra 565/deploy to release repo

jorntx:
First _very basic stab_ at establishing a release pipeline through our internal nexus.

You will need a full refresh from aegis and it is also there you will find the release instructions.


jesperlauridsen (approved):
Looks good!

PR #164, opened 2024-05-16
Configurable partner id

jorntx:
Very small change to make partner id configurable


jesperlauridsen (approved):
Makes sense. Also, I can see its already added on the config side.

PR #166, opened 2024-06-20
DRA-834 - New creator_affiliation_facet.

jesperlauridsen:
Now much to it. Changed the creator_affiliation to _facet where it needed to. And changed the few placed we used the first element of the array to just use the generic name in creator_affiliation now.


jorntx (approved):
Straightforward and works

PR #167, opened 2024-06-21
Add new base url and new confs for video and audio players

jorntx:
As the title says and remember fresh conf from Aegis.

Note: streaming maybe unavailable due to backend maintenance and fallouts at cloud vendor


jesperlauridsen (approved):
Looks good.

PR #169, opened 2024-06-24
DRA-bugfix: Add check for thumbnail behavior

jesperlauridsen:
We missed a check for the mounted method of getting thumbnails from Kaltura. Since we reuse the components, sometimes coming from a video to a audio result would add a bad request for a thumbnail from kaltura. This minor fix fixes this.


jorntx (approved):
Looks good

PR #170, opened 2024-06-25
DRA-bugfix - fix placement of arrow on fullpost

jesperlauridsen:
This makes sure the back arrow aligns with the text.


jorntx (approved):
This seems to be OK but it has all the slider stuff with it, so lets keep our hands of the merge button for now.

jorntx (approved):

PR #171, opened 2024-06-25
DRA-821 - background color on autocomplete

jesperlauridsen:
I'm not quite sure why this was a bug in some way - but I've added the white background color to both the ul list and the li list elements. Now it should be *almost* impossible to have a transparent background, no matter what browser and how it decides to interpret the code.


jorntx (approved):
Check!

PR #173, opened 2024-06-25
DRA-245 - safari issues fixed

jesperlauridsen:
Should fix weird positioning of elements on the frontpage and the font problem


jorntx (approved):
Seems to take of it - no funny business registered in my Safari

PR #174, opened 2024-06-26
DRA-915 - spinner needs to go away after failed solr query.

jesperlauridsen:
This should make sure even if we fail, we sent an error and set the result to zero which promts the no hits page. The error needs redesign, but that's part of a larger task still at Daniels desk.


jorntx (approved):
Works like a charm!

PR #176, opened 2024-07-02
Added first skeleton for 404. We didn't have  that i realized.

jesperlauridsen:
It occured to me we didn't even have the template for a 404 - so here it is!


jorntx (approved):
Thanks for the skeleton - works as intended

PR #177, opened 2024-07-03
Add timer and refresh function for the user session

jorntx:
Very straight forward and small implementation until we get real OAUTH going.

Set refreshSessionBufferTime to 0 and sessionTime to 5000 or something else low to test it out. You should see a call to refresh/authenticate according to the time outs you set.


jesperlauridsen:
Tested with some low values, and it seems to work very nicely - when looking at the networks tab i see a lot of authentication requests. Nice!

jesperlauridsen (approved):
\o/

PR #178, opened 2024-07-05
DRA 855 -  Adding timesearch to facets

jesperlauridsen:
--- DRAFT-MODE ---

So, this is a rough first draft. Most of the basic stuff works - still needed:

If the expanded searchbar is open, and you do a new query search, we need to apply the time filters to the search. Doesn't do that right now.

Add t he possiblity to change to date sort in the sort menu. Also translation of the search.startTime and title_sort_da variable as a filter

Some seriously thorough testing!

Will do these things when i get back. For now, this is pretty solid i think

### DIFF
--- a/src/components/search/Facets.vue
+++ b/src/components/search/Facets.vue
@@ -1,19 +1,12 @@
 <template>
 	<div
 		ref="facetsContainer"
-		class="search-facets active"
+		class="search-facets"
 	>
-		<div class="facet-background">
-			<button
-				class="facet-close-button"
-				@click="searchResultStore.toggleShowFacets(false)"
-			>
-				<span class="material-icons">close</span>
-			</button>
-
-			<div class="facet-container">
-				<div>
-					<h2 class="headline">{{ t('search.channels') }} ({{ currentFacetNr }})</h2>
+		<div class="facet-container">
+			<div class="flex-container">
+				<h2 class="headline">{{ t('search.channels') }} ({{ currentFacetNr }})</h2>
+				<div class="facet-options">
 					<TransitionGroup name="result">
 						<div
 							v-for="(singleFacet, index) in currentFacetNr as unknown as FacetPair[]"
@@ -29,6 +22,11 @@
 							/>
 						</div>
 					</TransitionGroup>
+				</div>
+			</div>
+			<div class="flex-container">
+				<h2 class="headline">{{ t('search.relatedSubjects') }} ({{ categoryNr }})</h2>
+				<div>
 					<CategoryTags
 						:categories="categoryFacets"
 						:category-nr="categoryNr"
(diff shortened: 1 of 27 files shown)


jorntx (commented):
Thanks for the sneak peak through this draft PR.

Looks good and heading in the right direction by splitting TimeFilters out and building Facets with TimeSearchComponent that includes TimeFilters.

So far I only found minor stuff (translation, console.logs etc.) that probably will disappear once you submit the final PR.

src/components/search/Facets.vue
```diff
@@ -105,14 +129,87 @@ export default defineComponent({
...
 		);
 
+		/* 		watch(
+			[months, days, timeslots, timeSliderValues],
+			() => {
+				console.log('SOMETGING HAPPENED');
+				searchResultStore.getSearchResults(searchResultStore.currentQuery, true);
+			},
+			{ deep: true },
+		); */
```
jorntx, line 139: I guess this can be deleted?

src/locales/en.json
```diff
@@ -57,15 +65,15 @@
...
      },
      "timeslots": {
-         "morning":"morning",
-         "midday":"midday",
-         "evening":"evening",
-         "night":"night"
+      "morning":"06.00 - 12.00",
+      "midday":"12.00 - 18.00",
+      "evening":"18.00 - 24.00",
+      "night":"00.00 - 06.00"
```
jorntx, line 71: I guess you need some am/pm indication here instead?

jorntx (commented):
Thanks for the changes from the draft PR - looks very good and only minor comments to the code.

Overall 1: As discussed we need to do something about the fact that the 'time machine widgets' take up so much screen real estate that you don't notice that the search results live update. At least some kind of spinner/hint near the widgets should be introduced but an even better solution IMHO would be to compact the 'time machine widgets' 

Overall 2: If you use the time machine and fold out/use _all the widgets_ and and perform a 'see all results' you will have to fold out all the widgets again to perform an adjustment.

Bug: On the front page and the search result page - if you first use the time slider and then toggle the drop down the slider does not update.

Something to discuss: If you use the 'time machine' and perform a 'see all results' and then use the back button then all is lost and you most start over..

src/components/common/CustomExpander.vue
```diff
@@ -0,0 +1,257 @@
...
+			>
+				{{ expanderOpen ? '-' : '+' }}
+			</button>
+		</div>
+		<div class="dotted-line"></div>
+	</div>
+</template>
+
+<script lang="ts">
+import { onUnmounted } from 'vue';
```
jorntx, line 28: Remove unused `onUnmounted`

src/components/common/CustomExpander.vue
```diff
@@ -0,0 +1,257 @@
...
+import { defineComponent, onMounted, ref, PropType } from 'vue';
```
jorntx, line 29: Remove unused `onMounted`

src/locales/en.json
jesperlauridsen: Right now it's 24 our interval - should be enough no? Actually I'm not really fuzzed about it - i would rather throw this to the UX/Content people.

jorntx (approved):
Yes, we are good to go!

PR #182, opened 2024-08-16
Rewire endpoints to go through bff proxy to enforce OAUTH

jorntx:
New paths as we are now live with OAUTH!


jesperlauridsen (approved):
Works! Woohoooo!

PR #183, opened 2024-08-16
Refactor to use entryId instead of refrenceId when streaming audio and video

jorntx:
The title says it all - small PR. Only way to test is on devel.


jesperlauridsen:
This looks very nice! - An empty string if none is found, and then the player isn't even loaded. Good job \o/

jesperlauridsen (approved):

PR #184, opened 2024-08-26
DRA-853: reworked checkbox for channels

jesperlauridsen:
Okey, this is slightly bigger than it should have been - but Daniel kept coming back with more changes. Here's the overview:

Facets have now been completely rebuild - this PR and the one before took care of this. In this PR, the following is done:

- Channel facets now has a new layout, and has no limit. So we get all of them.
- Channel facets now only change on time and search term changes. This change was requested by UX. So you can now pick several channels on a single overall query, and see them displayed. We don't change them if the query is only updated by a channel facet. This meant a lot of work and a lot of checks for what is actually requested - and a lot of the PR is just that.
- All filters in a container now has the the selected displayed next to the headline - as buttons that can be removed again.
- This also means that the seprated facet request for the backend no longer has other applied channel facets send with it.
- Moved facets and search result headline around - requsted from UX.
- Added a reset filters - requested from UX (naming not done yet - UX is still figuring this stuff out)
- Added a "Luk tidspunkter" button at the bottom of the time search filters - requested by UX.
- Removed the months time selectors on the filter options. Requested by UX.

Overall, I think this works nicely. But a good rigorous testing would be lovely - there are now a lot of moving parts here now, when we change query to make sure we do things the right way.

### DIFF
--- a/src/components/search/SearchOverhead.vue
+++ b/src/components/search/SearchOverhead.vue
@@ -176,18 +176,12 @@ export default defineComponent({
 			}
 			searchResultStore.preliminaryFilter = val;
 			searchResultStore.resetAutocomplete();
-			let query: LocationQueryRaw = {
-				q: searchResultStore.currentQuery,
-				start: 0,
-			};
-			if (searchResultStore.preliminaryFilter !== '') {
-				query.fq = searchResultStore.preliminaryFilter;
-			}
-			if (searchResultStore.sort !== '') {
-				query.sort = searchResultStore.sort;
-			} else {
-				query.sort = encodeURIComponent(`score desc`);
-			}
+			const query = { ...route.query };
+			console.log(route.query);
+			let filters = [] as string[];
+			query.fq !== undefined ? (filters = [...(query.fq as string[])]) : null;
+			filters = filters.filter((str) => !str.includes('origin'));
+			query.fq = filters;
 
 			router.push({
 				name: 'Home',
--- a/src/components/search/SearchOverhead.vue
+++ b/src/components/search/SearchOverhead.vue
@@ -77,7 +77,6 @@
 							:query="searchResultStore.lastSearchQuery !== undefined ? searchResultStore.lastSearchQuery : ''"
 						/>
 					</div>
-					<!-- <Sort v-if="searchResultStore.searchResult.length > 0" /> -->
 				</div>
 				<div class="sort-options">
 					<Sort></Sort>
--- a/src/components/search/SearchOverhead.vue
+++ b/src/components/search/SearchOverhead.vue
@@ -2,30 +2,29 @@
 	<div class="container">
 		<div class="row">
 			<div class="hit-count">
-				<div class="result-options">
-					<div class="hits">
-						<HitCount
-							:hit-count="searchResultStore.numFound"
-							:no-hits="searchResultStore.noHits"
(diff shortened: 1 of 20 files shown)


src/components/search/SearchOverhead.vue
```diff
@@ -70,6 +69,16 @@
...
 				<Facets :facet-results="searchResultStore.facetResult" />
+				<div class="result-options">
+					<div class="hits">
+						<HitCount
+							:hit-count="searchResultStore.numFound"
+							:no-hits="searchResultStore.noHits"
+							:query="searchResultStore.lastSearchQuery !== undefined ? searchResultStore.lastSearchQuery : ''"
+						/>
+					</div>
+					<!-- <Sort v-if="searchResultStore.searchResult.length > 0" /> -->
```
jorntx, line 80: Lets get rid of this dead code

src/components/search/SearchOverhead.vue
```diff
@@ -163,35 +172,72 @@ export default defineComponent({
...
 
+		watch(
+			() => route.query.fq,
+			(newFq) => {
+				/* 
+				we have to do this, because vue acts weird here.
+				when we have normal route changes, it seems like we get an array here,
+				but when we use brower back and forth buttons, we get strings, IF there is only one filter.
+				Weird and breaking behavior, that we have to account for here.
+				*/
```
jorntx, line 183: ouch!

src/store/timeSearchStore.ts
```diff
@@ -13,17 +13,33 @@ export const useTimeSearchStore = defineStore('timeSearchStore', () => {
 	let comparisonSearchUUID = '';
 	const searchFired = ref(false);
 	const numFound = ref(0);
+	/* these are just the preset values. Should be updated when we got the new repo */
+	const startYear = ref(1921);
+	const endYear = ref(2022);
```
jorntx, line 18: Is it possible to move the values to the timesearch config?

src/utils/filter-utils.ts
```diff
@@ -57,6 +63,80 @@ const removeFilter = (route: RouteLocationNormalizedLoaded, filter: string) => {
...
+		} else {
+			existingFq.push(encodeURIComponent(`${newFilter}`));
+		}
+		routeQueries.fq = existingFq;
+	} else {
+		routeQueries.fq = [];
+		routeQueries.fq.push(encodeURIComponent(`${newFilter}`));
+	}
+	return routeQueries;
+}
```
jorntx, line 92: Looot of action here - can you comment or maybe break the method up into smaller chunks?

src/utils/filter-utils.ts
```diff
@@ -57,6 +63,80 @@ const removeFilter = (route: RouteLocationNormalizedLoaded, filter: string) => {
...
+			filters = filters.filter((filter) => filter !== encodeURIComponent(filterToRemove));
+			const finalFilter = `(${filters.join(' OR ')})`;
+			if (filters.length !== 0) {
+				existingFq.push(finalFilter);
+			}
+		}
+		routeQueries.fq = existingFq;
+	}
+	return routeQueries;
+}
```
jorntx, line 117: See my comment above

jorntx (commented):
Very nice work - left some comments for you to look over and then I will give it a spin later

jorntx (approved):
Thanks for the adjustments - lets roll!

PR #187, opened 2024-09-04
Dra 1083/handle auth errors

jorntx:
This took a little while to figure out how to handle this "authentication" business. Take a look at the code and then we can talk about how we test this. 

In short this PR handles the 401 scenario and auto refresh of the user session:
    
    - handle 401 in api-client
    - add global service retry logic in api-client if service calls encounter 401
    - retry 2 times if 401 occours and auth call fails
    - set session to 36 hours to match backend
    - refresh session 15 minutes before session expires


src/store/authStore.ts
```diff
@@ -3,44 +3,53 @@ import { APIService } from '@/api/api-service';
...
+	 *  refresh will run 15 minutes before session espires
+	 */
+	const refreshSessionBufferTime = 900000;
 
 	let refreshTimeout: ReturnType<typeof setTimeout>;
+	const isAuthenticating = ref(false); // Lock to prevent multiple authentication calls
 
 	const authenticate = async () => {
+		// Check if authentication is already in progress (omitted the if brackets on purpose here:) )
+		if (isAuthenticating.value) return;
```
jesperlauridsen, line 31: Ooooh, different standards for _senior_ developers, eh?!

src/api/api-client.ts
```diff
@@ -38,17 +40,72 @@ export class APIServiceClient {
...
+						await this.handleAuthenticationWithRetry(authStore);
+
+						/** A bit complex here but what is happening is in short:
+						 *  Retry all pending requests after successful authentication (thats all
+						 * 	the map((req) => req() business) and subsequently clear the queue.
+						 *  After that we give the original request a go.
+						 * 	If the auth call fails, we clear the pending requests queue and reject
+						 *  with an error.
+						 */
+						await Promise.all(this.pendingRequests.map((req) => req()));
```
jesperlauridsen, line 81: This is some complex stuff!

jesperlauridsen:
This *looks* solid. I think we should put our heads together tomorrow and see if there's a good way to test it - or you might already have done that? Either way, code looks good, albeit a little complex at times.

src/store/authStore.ts
jorntx: Wuff, that stings and I remember now :) I don't think there should be double standards - I will fix. And btw. you too are a senior now in my book ;)

src/api/api-client.ts
jorntx: I totally agree and actually my first pass was more verbose and maybe clearer. I will see if I can do anything about the complexity.

jesperlauridsen (approved):
This looks very solid. Looking forward to testing this on dev.

PR #188, opened 2024-09-04
DRA-953: Accessability fixes

jesperlauridsen:
Some accessability fixes here and there:

- Did a npm audit fix - removed some errors and updated the vue3-slider-component to handle some issues.
- Fixed a few issues with the slider. It now has tabindex -1 so we stick to the dropdowns if you're navigating with a keyboard.
- Reworked the checkboxes a little, to have a legend and fieldsets. A little tricky because of our structure, but my Linux screenreader is okey with this for now, as is Lighthouse. Gonna be fun to see what Monsido thinks of it.
- Found a small issue with our images and links in resultitem.
- Found a small issue in our menu - easy fix to make lighthouse happy.

### DIFF
--- a/src/components/common/CustomTimelineSelect.vue
+++ b/src/components/common/CustomTimelineSelect.vue
@@ -7,12 +7,12 @@
 				@change="updated($event)"
 			>
 				<option
-					value="1992"
+					:value="startTime"
 					disabled
 					selected
 					hidden
 				>
-					1992
+					{{ startTime }}
 				</option>
 				<option
 					v-for="(item, index) in listItems"
@@ -30,6 +30,7 @@
 
 <script lang="ts">
 import { defineComponent, PropType, ref, onMounted, watch } from 'vue';
+import { initStartDate } from './TimeSearch/TimeSearchInitValues';
 
 export default defineComponent({
 	name: 'CustomTimelineSelect',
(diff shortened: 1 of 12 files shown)


src/components/search/ResultItem.vue
```diff
@@ -165,7 +166,7 @@ export default defineComponent({
 		const getImageData = () => {
 			const imageDataObj = {} as ImageComponentType;
 			imageDataObj.altText = props.resultdata?.title;
-			imageDataObj.imgTitle = props.resultdata?.title;
+			imageDataObj.imgTitle = props.resultdata?.title ? props.resultdata.title : 'se materiale';
```
jorntx, line 169: No translation needed here?

src/components/common/CustomTimelineSelect.vue
```diff
@@ -1,26 +1,30 @@
...
+			{{ label }}:
+			<select
+				v-model="selected"
+				@change="updated($event)"
 			>
-				{{ item }}
-			</option>
-		</select>
+				<option
+					value="1992"
```
jorntx, line 10: Wouldn't it be better to fetch this value from TimeSearchInitValues?

src/components/common/CustomTimelineSelect.vue
```diff
@@ -1,26 +1,30 @@
...
+					disabled
+					selected
+					hidden
+				>
+					1992
```
jorntx, line 15: Wouldn't it be better to fetch this value from TimeSearchInitValues?

jorntx (commented):
I'm happy with the changes and app still works but as you mentioned Monsido really has the last word here.

Only have a few comments and then we are ready to merge.

src/components/search/ResultItem.vue
jesperlauridsen: Done!

jorntx (approved):
Looks good!

PR #190, opened 2024-09-05
DRA-1245 - deselect bug in timesearch check all button

jesperlauridsen:
Very simple. When you use a select all in either of the time search checkboxes (months, days, timeslots), when you press the all button, it should select all, or deselect all. Before, when you deselected, the first one would always be set to true aswell. Now they all deselect, as UX want it now (as reported in DRA-1245)


jorntx (approved):
Fix verified and testet OK

PR #193, opened 2024-09-05
OPTIONAL: Small fix to make the top bar act appropriately

jesperlauridsen:
This is just a small fix to get the header more in line with the design of kb.dk when on a page that's using the cutout. Very simple not much to it. No JIRA on this, i just found it while fixing something else and decided to quickly do it too.


jorntx (approved):
Seems to work as before minus the changes. Removed a console.

PR #195, opened 2024-09-10
small fix to make them not-so-unique. Requested by testers

jesperlauridsen:
Super simple - they needed them to not be tied to the post, but rather just to the number in the list. Quick fix.


jorntx (approved):
If the testers say so:)

PR #197, opened 2024-09-16
Fix infinite retry loop triggered by http 401

jorntx:
Fix infinite loop triggered by consistent 401 response from bff. Fixed by implementing a retry threshhold on pending requests.


jesperlauridsen (approved):
I understand it, and it looks solid. Looking forward to testing this on devel!

PR #203, opened 2024-10-10
First stab at genre facets

jesperlauridsen:
This PR takes care of a few different things, but mainly this builds categorys (or genre), into the facets.

Wrapped calls for facets for the initial build into one call, which then waits for our authentication.
Build up the genre facets for the search results, with icons and whatnot.
Slight brushup on the custom expander, designwise.
Failsafe for if we don't get results from the backend for the slider.
reworked the addChanneFilter and removeChannelFilter so they can be used for both genre and channel


src/views/Search.vue
```diff
@@ -165,10 +160,16 @@ export default defineComponent({
...
 				fq.includes('creator_affiliation_facet'),
 			);
+			const newGenreFilter = (newFilter as string[])?.find((fq: string) => fq.includes('genre'));
+			const oldGenreFilter = (oldFilter as string[])?.find((fq: string) => fq.includes('genre'));
+
 			if (newCreatorAffiliationFilter !== oldCreatorAffiliationFilter && newfq.query.q === oldfq.query.q) {
-				return true;
+				searchResultStore.setKeepChannels(false);
+				searchResultStore.setKeepGenre(false);
+			} else if (newGenreFilter !== oldGenreFilter && newfq.query.q === oldfq.query.q) {
```
Pbruun, line 169: Is this to avoid a more complex if statement?

jesperlauridsen: Good point! Fixed!

jesperlauridsen:
Self-reviewed, and reviewed by Patrick - and merged.

PR #205, opened 2024-10-21
Dra 881 implement 404 page

Pbruun:
404 page when url doesn't exists.
contains kb 404 image. Buttons for the frontpage of dr-archive and kb.dk
Displays content differently when page gets smaller.
test id for link buttons

### DIFF
--- a/src/views/404.vue
+++ b/src/views/404.vue
@@ -19,7 +19,7 @@
 						class="btn-light btn"
 						href="/find-materiale/dr-arkivet/"
 						role="button"
-						:data-testid="addTestDataEnrichment('button', '404', 'link-to-home')"
+						:data-testid="addTestDataEnrichment('button', '404', 'link-to-home', 0)"
 					>
 						{{ t('error.wrongUrl.frontPage') }}
 					</a>
@@ -27,7 +27,7 @@
 						class="btn-light btn"
 						href="https://www.kb.dk"
 						role="button"
-						:data-testid="addTestDataEnrichment('button', '404', 'link-to-kb')"
+						:data-testid="addTestDataEnrichment('button', '404', 'link-to-kb', 0)"
 					>
 						{{ t('error.wrongUrl.kbPage') }}
 					</a>
@@ -176,6 +176,7 @@ export default defineComponent({
 /* MEDIA QUERY 990 */
 @media (min-width: 990px) {
 	.not-found-content {
+		justify-content: space-between;
 		display: flex;
 		max-width: 1150px;
 		flex-direction: row;
--- a/src/views/404.vue
+++ b/src/views/404.vue
@@ -19,6 +19,7 @@
 						class="btn-light btn"
 						href="/find-materiale/dr-arkivet/"
 						role="button"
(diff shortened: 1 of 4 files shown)


src/views/404.vue
```diff
@@ -1,9 +1,38 @@
...
+			<div class="not-found-details">
+				<h1>{{ t('error.wrongUrl.header') }}</h1>
+				<p>{{ t('error.wrongUrl.text') }}</p>
+				<h2>{{ t('error.wrongUrl.altHeader') }}</h2>
+				<div>
+					<a
+						class="btn-light btn"
+						href="/find-materiale/dr-arkivet/"
+						role="button"
+						:data-testid="addTestDataEnrichment('button', '404', 'link-to-home')"
```
jesperlauridsen, line 22: This is silly - but the function addTestDataEnrichment actually _requires_ for arguments. The last probably should be optional, but for now can you add a 0 as the fourth argument here, so, just so its consistent with the rest of the test id's throughout the site?

src/views/404.vue
```diff
@@ -1,9 +1,38 @@
...
+					>
+						{{ t('error.wrongUrl.frontPage') }}
+					</a>
+					<a
+						class="btn-light btn"
+						href="https://www.kb.dk"
+						role="button"
+						:data-testid="addTestDataEnrichment('button', '404', 'link-to-kb')"
```
jesperlauridsen, line 30: Same as the above.

src/views/404.vue
```diff
@@ -31,10 +67,134 @@ export default defineComponent({
...
+}
+/* MEDIA QUERY 640 */
+@media (min-width: 640px) {
+	.not-found-content {
+		max-width: 990px;
+	}
+}
+/* MEDIA QUERY 990 */
+@media (min-width: 990px) {
+	.not-found-content {
```
jesperlauridsen, line 178: Adding justify-content: space-between; here will make the 2 divs align with the outer edges of the container. Then we can always introduce padding on the sides if we want it to align slightly further inwards instead.

src/views/404.vue
```diff
@@ -12,15 +41,22 @@
...
 export default defineComponent({
 	name: '404',
 	components: {
 		Footer,
 	},
 	setup() {
-		return {};
+		const { t } = useI18n();
+		const getImgServerSrcURL = () => {
+			return new URL(`@/assets/images/404.svg`, import.meta.url).href;
```
jesperlauridsen, line 55: You're looking for an image here that doesn't exist. Did you forget to commit it? It's on in the PR and i get a broken image icon on the website when I try it out :o

jesperlauridsen:
Much better, now the image is there! However, I've noticed that on big screens (1280 or higher, which most of us use these days), the content doesn't align properly with the width of the other elements (in this case the searchfield is a good indicator of the width of the containers)

image


To fix this, that best way would probably be to have an outer container with width 100% or 100vw and display:flex and align-items: center; / justify-content: center; and the background-color, and then an inner container that follows the max-widths set from the media queries.

PR #207, opened 2024-10-22
DRA-1380: Fullwidth background on the time facets

jesperlauridsen:
This adds a full-width background to the time filters when opened.
Minor tweaks in a few other places to make sure this doesn't break anything on the frontpage either


Pbruun:
Looks good!
Only thing I noticed was that the content of pick weekday is a bit more to the right than the content of pick month.

PR #208, opened 2024-10-29
Dra 1400 text changes

Pbruun:
Text changes according to UX user test. 

Some styling needed on text in noHits. Will be done on noHits branch.


src/components/search/NoHits.vue
```diff
@@ -9,20 +9,39 @@
...
 			</div>
 			<TiltedDivider
-				title="Et par fif til din søgning"
+				title="Gode råd til din søgning"
 				:right="true"
 			></TiltedDivider>
 			<TextAndImage>
 				<template #text>
 					<div>
-						<h3>Her får du et par fif til at optimere din søgning</h3>
```
jesperlauridsen, line 18: We would want this in the translation files.

src/components/search/NoHits.vue
```diff
@@ -9,20 +9,39 @@
...
+						<h3>Måske kan du forbedre din søgning med en af disse metoder</h3>
```
jesperlauridsen, line 18: We would want this in the translation files.

src/components/search/Facets.vue
```diff
@@ -89,7 +89,7 @@
 								</CustomExpander>
 							</div>
 							<CustomExpander
-								headline="Udvælg kanaler"
```
jesperlauridsen, line 92: We would want this in the translation files.

src/components/search/Facets.vue
```diff
@@ -56,7 +56,7 @@
 						<div class="flex-container">
 							<div class="category-container">
 								<CustomExpander
-									headline="Udvælg kategorier"
```
jesperlauridsen, line 59: We would want this in the translation files.

src/components/search/AdditionalInfo.vue
```diff
@@ -8,7 +8,7 @@
 			@click="showThumbnails()"
 		>
 			<span class="material-icons thumbnails-icon">photo_library</span>
-			Thumbnails
```
jesperlauridsen, line 11: We would want this in the translation files.

src/components/common/timeSearch/TimeSearchFilters.vue
```diff
@@ -32,12 +32,12 @@
 					class="explanation-for-data"
 				>
 					<div>
-						Datamængde viser en kurve over data vi har i arkivet lorem ipsum dolor sit amet, consectetur adipiscing
-						elit, sed do eiusmod tempor incididunt ut.
+						Kurven viser, hvor meget DR-arkivmateriale du kan søge i fra de forskellige årstal. I nogle perioder er der
+						meget materiale, og i andre er der kun lidt eller ingenting.
 					</div>
 					<div>
-						Der er huller i arkivet, men vi arbejder på det lorem ipsum dolor sit amet, consectetur adipiscing elit, sed
```
jesperlauridsen, line 39: We would want this in the translation files.

src/components/common/timeSearch/TimeSearchFilters.vue
```diff
@@ -32,12 +32,12 @@
 					class="explanation-for-data"
 				>
 					<div>
-						Datamængde viser en kurve over data vi har i arkivet lorem ipsum dolor sit amet, consectetur adipiscing
```
jesperlauridsen, line 35: We would want this in the translation files.

src/components/common/PortalContent.vue
```diff
@@ -33,12 +33,12 @@
 		</div>
 		<div class="time-search">
 			<SkewedFoldable
-				:title="'Brug tiden...'"
+				:title="'Søg via periode'"
 				:left="false"
 				bg="#fafafa"
 				text="#0a2e70"
 				icon="schedule"
-				:subtitle="`Vælg, hvilket tidsrum i DR's sendeflader, du vil søge i.`"
```
jesperlauridsen, line 41: We would want this in the translation files.

src/components/common/PortalContent.vue
```diff
@@ -33,12 +33,12 @@
 		</div>
 		<div class="time-search">
 			<SkewedFoldable
-				:title="'Brug tiden...'"
```
jesperlauridsen, line 36: We would want this in the translation files.

src/components/search/NoHits.vue
```diff
@@ -9,20 +9,39 @@
...
 						<ul>
-							<li>Et område afsat til at lære brugerene om brugen af søgefeltet</li>
-							<li>Hvad kan man og ikke mindst hvad man kan ikke søge på</li>
-							<li>Vil det være muligt at søge på en titel</li>
-							<li>Vil det være muligt at søge på en programrække</li>
-							<li>I det hele taget skal der laves en "søgeguide"</li>
-							<li>Her skal stå hvordam man bruger vores søgefelt</li>
+							<li>
```
jesperlauridsen, line 20: We should have this in the translation files. you could do it something like this:

searchHelp: {
first: "",
second:"",
third:"",
},

and then plug them in, in the divs. Would we good so we can get them translates aswell.

jesperlauridsen (changes requested):
This is some very nice work. We _try_ not to have any direct text in the code, so all the places where something is written, should be in the translation files.

src/components/common/PortalContent.vue
```diff
@@ -2,12 +2,10 @@
 	<div class="portal-container container">
 		<div class="categories">
 			<SkewedFoldable
-				:title="'Kategorier'"
```
jesperlauridsen, line 5: The complete removal of a title here is unfortunate. Removing it completely right now causes the component to just use "Title" here. That's wrong, and was never intended to happen, but is easily fixable

image


Secondly, and more important, on mobile these titles are crucial for the funcitonality and understanding of the components. Where is the screenshot from devel:

image

I'm quite certain that they're needed. The component takes into account if they should be shown on desktop, but I do believe that need to be added (both title and subtitle actually), for the mobile design to work. (They're also still in the mobile design UX file.)

src/components/common/PortalContent.vue
```diff
@@ -2,12 +2,10 @@
...
 				:left="false"
 				bg="#002e70"
 				text="white"
 				icon="category"
-				:subtitle="'Vælg, hvilken kategori du gerne vil søge i.'"
```
jesperlauridsen, line 10: Same as the above :-)

jesperlauridsen (approved):
Very good job! Approved!

PR #210, opened 2024-10-31
Dra 1349 no result page

Pbruun:
Displays all the suggestions for the search. 
Got some advice for a better search


src/components/global/content-elements/TextAndImage.vue
```diff
@@ -37,4 +42,17 @@ temporary styling until patterns from design system are implemented
 	justify-content: center;
 	flex-wrap: wrap;
 }
+@media (min-width: 450px) {
+	.imageAndTextContainer {
+		flex-direction: column;
+	}
+	.imageAndTextContainer .con {
+		width: 80%;
```
jesperlauridsen, line 50: Why 80%? This makes it not centeret on a small screen, which seems a little odd?

src/locales/en.json
```diff
@@ -157,7 +167,10 @@
            "line1": "Try writing * at the end of a word, for example skov*",
            "line2": "Then you'll search for skov, skovens, skovtur and so on"
         },
-        "fourth": "Use searching via date or periode om DR-archive's frontpage, if you got a sense of where the broadcast was sent"
+        "fourth": "Use searching via date or periode om DR-archive's frontpage, if you got a sense of where the broadcast was sent",
+        "readMore": "Read more about",
+        "searchTips": "search tips",
+        "link": ""
```
jesperlauridsen, line 173: We need to remember this some way - Maybe add a JIRA tastk to get this updated before launch?

jesperlauridsen (changes requested):
This seems very good! I have a few comments, but other than that i got nothing for the code. I got a few concerns about the design, but those do not belong here. Gj, have a look at the 2 comments i left, and when you're good to go I'll merge this \o/

jesperlauridsen:
Also, 
image
If you do a combination of a search, and some selections, in this case http://localhost:3000/find-materiale/dr-arkivet/find?q=bl%C3%A5&start=0&sort=score+desc&fq=creator_affiliation_facet%253A%2522DR%2520P3%2522&fq=genre%253A%2522Videnskab%2520og%2520natur%2522

 a saerch on "blå", and a selected creator (DR P3) and a genre (videnskab og natur), you get nothing - but the "mente du" shows up, but its empty. That should either contain something, or not be present, no?

src/components/global/content-elements/TextAndImage.vue
Pbruun: I made it 80%, because i thought the line breaks seemed a bit odd. 
Changed to 100% so it is center.

Pbruun:
Link to "search tips" is now correct on the danish page. English version will be made later today. 
We had multiple spellings of "DR-archive" UX asked for it to be made similar.

PR #212, opened 2024-11-07
Dra 864 detail display

Pbruun:
Changed styling according to UX with display of record details.
Video/ audio player works on stage, so doesn't seems to need changes.


src/components/records/BroadcastVideoRecord.vue
```diff
@@ -14,7 +14,7 @@
 		<div class="boardcast-record-data">
 			<div class="main-record-data">
 				<div class="record-data">
-					<h2>{{ recordData.name[0].value }}</h2>
+					<h2>{{ recordData.name[0].value ? recordData.name[0].value : recordData.name }}</h2>
```
jesperlauridsen, line 17: You changed this here nicely - but not in BroadcastAudioRecord. Is there a reason for this? Also, in BroadcaastAudioRecord, we still have this:

<div class="record-data">
					<h2>{{ recordData.name }}</h2>
					<p>{{ recordData.description }} ALSO, LOREM IPSUM PLEASE</p>
				</div>
				
				
				I think we can remove the "ALSO, LOREM IPSUM PLEASE" part now :)

src/components/records/BroadcastVideoRecord.vue
```diff
@@ -44,14 +44,19 @@
...
-					<span class="material-icons">link</span>
-					<span class="link-text">{{ $t('record.copy') }}</span>
-				</button>
+				<div class="share-button">
+					<div
+						class="link-container get-link"
+						:data-testid="addTestDataEnrichment('button', 'broadcast-video', 'copy-link', 0)"
+						@click="getCurrentUrl()"
+					>
+						<img
```
jesperlauridsen, line 53: This icon is from the material icons, so we would much rather use the font for this:

<span class="material-icons offset">share</span> would do the trick, without having to load an svg and all that jazz.

src/components/records/BroadcastAudioRecord.vue
```diff
@@ -43,14 +43,19 @@
...
-					<span class="material-icons">link</span>
-					<span class="link-text">{{ $t('record.copy') }}</span>
-				</button>
+				<div class="share-button">
+					<div
+						class="link-container get-link"
+						:data-testid="addTestDataEnrichment('button', 'broadcast-video', 'copy-link', 0)"
+						@click="getCurrentUrl()"
+					>
+						<img
```
jesperlauridsen, line 52: This icon is from the material icons, so we would much rather use the font for this:

<span class="material-icons offset">share</span> would do the trick, without having to load an svg and all that jazz.

src/components/records/BroadcastAudioRecord.vue
```diff
@@ -146,7 +151,9 @@ export default defineComponent({
 		const getCurrentUrl = () => {
 			copyTextToClipboard();
 		};
-
+		const getImgServerSrcURL = () => {
+			return new URL(`@/assets/images/share.svg`, import.meta.url).href;
```
jesperlauridsen, line 155: We can remove this and just use the material icons font

src/components/records/BroadcastVideoRecord.vue
```diff
@@ -143,7 +148,9 @@ export default defineComponent({
 		const getCurrentUrl = () => {
 			copyTextToClipboard();
 		};
-
+		const getImgServerSrcURL = () => {
```
jesperlauridsen, line 151: We can remove this and just use the material icons font

jesperlauridsen (changes requested):
Very nice work. I got a few minor comments for you to look over.

src/components/records/BroadcastVideoRecord.vue
Pbruun: The Lorem Ipsum part has been changed in another branch.
I didn't change in BroadcastAudioRecord because it just had recordData.name. But I'll change it, just to be sure.

jesperlauridsen (approved):
Good job, looks good!

PR #213, opened 2024-11-10
DRA-optional: Bug-fixing

jesperlauridsen:
A few bugs found doing the tests (and a few i discovered myself)

- When searching from a fullpost, you now get a new set of results on the new search term.

- Doing a search from the "pick-a-date" component did not always get the filters attached to the query - this should be fixed now.

- Now picking a new month or year in the "pick-a-date" component sets the day to the first, so you don't _have_ to pick a date before doing a search.

- Fixed an error where you could get a weird breaking of the app because a response was not checked if existed before looking into the object.

- Fixed a bug where the values of the startDate and endDate were not reset when returning to the frontpage, thus messing up the timeline component's "go look at this"-link.

- Adjusted slider for new datapoints

- Minor change to titles for skewedFoldable components

- Adjusted start and end date on daypicker

- Fixed links for genre on fullpost

- Added tv/radio current filter options for visiblity

- Fixed a few bugs with reload and back/forth browser compatability

### DIFF
--- a/src/assets/styles/vue-slider-styles.css
+++ b/src/assets/styles/vue-slider-styles.css
@@ -124,17 +124,17 @@
 	}
 }
 
-.vue-slider-marks .vue-slider-mark:nth-child(5n) .vue-slider-mark-label {
+.vue-slider-marks .vue-slider-mark:nth-child(1n) .vue-slider-mark-label {
 	display: block;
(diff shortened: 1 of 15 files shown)


Pbruun:
Great work! Merging

PR #214, opened 2024-11-11
Dra 1334 related content by date

Pbruun:
If more like this doesn't give any results. It'll display content from the day. 
Styling to reuse GridResultItem component.


src/components/records/BroadcastVideoRecord.vue
```diff
@@ -61,31 +62,34 @@
...
+				<GridResultItem
 					:row-nr="3"
 					:spot-nr="3"
 					:draggable="true"
-					:spots="moreLikeThisRecords"
-				></GridDisplay>
+					:resultdata="record"
+					:starttime="new Date(record.startTime).toLocaleString()"
+					:index="index"
+					:full-post-url="true"
```
jesperlauridsen, line 90: The GridResultItem actually wants a :loading attribute awell, so it knows what state to display. Adding

:loading="moreLikeThisRecords.length === 0"

 will solve this i think, and remove the warnings in the console.

src/components/records/BroadcastAudioRecord.vue
```diff
@@ -60,31 +61,34 @@
...
 		<div class="extra-record-data">
 			<div
-				v-if="moreLikeThisRecords !== undefined && moreLikeThisRecords.length > 0"
+				v-for="(record, index) in moreLikeThisRecords"
+				:key="record"
 				class="related-content"
 			>
-				<h3>Relateret indhold</h3>
-				<GridDisplay
+				<GridResultItem
```
jesperlauridsen, line 82: The GridResultItem actually wants a :loading attribute awell, so it knows what state to display. Adding

:loading="moreLikeThisRecords.length === 0"

 will solve this i think, and remove the warnings in the console.

src/components/records/BroadcastVideoRecord.vue
```diff
@@ -61,31 +62,34 @@
...
-				Til forsiden
+				{{ $t('record.toFrontpage') }}
 			</router-link>
 		</div>
+		<h3>{{ $t('search.relatedContent') }}</h3>
 		<div class="extra-record-data">
 			<div
-				v-if="moreLikeThisRecords !== undefined && moreLikeThisRecords.length > 0"
+				v-for="(record, index) in moreLikeThisRecords"
+				:key="record"
```
jesperlauridsen, line 80: record is an object here, index is better as key, since you can be sure it's never the same.

src/components/records/BroadcastAudioRecord.vue
```diff
@@ -60,31 +61,34 @@
...
-				Til forsiden
+				{{ $t('record.toFrontpage') }}
 			</router-link>
 		</div>
+		<h3>{{ $t('search.relatedContent') }}</h3>
 		<div class="extra-record-data">
 			<div
-				v-if="moreLikeThisRecords !== undefined && moreLikeThisRecords.length > 0"
+				v-for="(record, index) in moreLikeThisRecords"
+				:key="record"
```
jesperlauridsen, line 79: record is an object here, index is better as key, since you can be sure it's never the same.

src/views/ShowRecord.vue
```diff
@@ -108,8 +114,27 @@ export default defineComponent({
...
+
+		const getStartAndEndFromStartTime = (): string[] => {
+			let startEnd: string[] = [];
+			if (recordData.value as BroadcastRecordType) {
+				const startTime = (recordData.value as BroadcastRecordType).startTime;
+				startEnd[0] = startTime.replace(/T(.*)/, 'T00:00:00Z');
+				startEnd[1] = startTime.replace(/T(.*)/, 'T23:59:59Z');
 			}
+			return startEnd;
 		};
```
jesperlauridsen, line 139: So, we have a problem here. If you press a new record, nothing happens, because we're already on that route, we gotta watch the params of the route and check for an id update, and then process that. Luckily, this page is build rather well (thanks to you), so this would do the trick:


		watch(
			() => route.params.id,
			() => {
				buildContentFromReponse().then(() => {
					window.scrollTo({ top: 0, behavior: 'smooth' });
				});
			},
		);

src/components/search/GridResultItem.vue
```diff
@@ -13,8 +13,8 @@
 				class="data"
 			>
 				<router-link
+					:to="{ path: fullPostUrl ? resultdata.id : 'post/' + resultdata.id }"
```
jesperlauridsen, line 16: I don't understand this. Why do we need this fullPostUrl? Is it because of a /post/post problem, because of the place of the route? If that's the case, a rebuild of the router-link like this:

:to="{
     name: 'Record',
     params: { id: resultdata.id },
}"

Would probably be better, then you can remove the extra fullPostUrl param you created.

jesperlauridsen (changes requested):
I have a few suggestions to changes here :-)

PR #215, opened 2024-11-11
Dra 167 not allowed

Pbruun:
A display if user is not allowed to view content. Contains text with why the content isn't playing, and a link to read more about dr-archive. 
With v-if="!isAllowedToDisplayContent" I haven't been able to test without changing a bit in NotAllowedBanner and then hotreload the page. Uses authStore.


src/components/global/content-elements/NotAllowedBanner.vue
```diff
@@ -0,0 +1,119 @@
...
+			</div>
+			<div class="not-allowed-lock">
+				<span class="material-icons">lock</span>
+			</div>
+		</div>
+	</div>
+</template>
+
+<script lang="ts">
+import { defineComponent, onMounted, PropType, ref } from 'vue';
```
jesperlauridsen, line 28: onMounted, PropType and ref are unused here
aswell is useRouter, useI18n

src/views/ShowRecord.vue
```diff
@@ -86,6 +89,7 @@ export default defineComponent({
 		const handleShowRecordError = (err: AxiosError, type: string) => {
 			switch (type) {
```
jesperlauridsen, line 90: Very nice use of a switch case here!

jesperlauridsen (changes requested):
I only had those few unused imports. Fix those and this is good to go!

PR #216, opened 2024-11-14
DRA-1469: reworked hover on day-picker on frontpage

jesperlauridsen:
Reworked the daypicker a little bit:*

Now has a hover on the entire element.
Now shows the fold out button on all layouts
Reworked, so the fold out button now has the same look as the other ones on the desktop layout.


Pbruun:
Looks great!

PR #217, opened 2024-11-14
Created component to display if user has no access to content.

Pbruun:
We needed a distinct component to display if user don't have access to the content. 
Cleaned up unused attributes.
Only mlt if we get a record.


jesperlauridsen (approved):
Seems to work nicely, gj

PR #220, opened 2024-11-15
Dra 1258 result sorting

Pbruun:
sort arrows. can flip between asc and desc. defaults to desc for each "category"


jesperlauridsen (approved):
Very nice work!

PR #221, opened 2024-11-18
show and reset buttons larger. tv and radio button larger

Pbruun:
Made the filter buttons larger (show, reset, tv, radio)


jesperlauridsen (approved):
Looks solid, gjgj!

PR #223, opened 2024-11-18
Dra 1470 return button

Pbruun:
Return button on record. Styling changed. It needed to be more visible. So its larger, and further up (follows description text)


jesperlauridsen (approved):
Seems to work! Gj!

PR #225, opened 2024-11-25
DRA-1412 HERO update

danielqvistcom:
Pkt. 1 #HERO update. Left/right coll. styling + font-updates.

### DIFF
--- a/src/components/search/Hero.vue
+++ b/src/components/search/Hero.vue
@@ -246,7 +246,7 @@ h1 {
 	display: block;
 	text-align: center;
 	text-decoration: none;
-	font-size: 26px;
+	font-size: 24px;
 	padding: 4px 10px 6px 10px;
 	border-radius: 4px;
 	height: 55px;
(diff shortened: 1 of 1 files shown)


jesperlauridsen:
On a small screen
image
the button font now overflows the container.

src/components/search/Hero.vue
```diff
@@ -436,7 +475,14 @@ h1 {
 		flex-direction: row;
 		max-width: 1150px;
 		justify-content: space-between;
-		align-items: center;
+	}
+	.container h1 {
+		margin-top: 105px;
+	}
+	.link-container .link:hover {
```
jesperlauridsen, line 482: Is there a reason for not putting a "transition: all 0.25s linear 0s" on this, to make the transition smoother?

jesperlauridsen: On this note, if you wanna animate the underline too, it should probably be a :after effect instead. text-decoration:underline has no animation features, so it just pops in and out visually.

danielqvistcom: You're right. Much smoother with the transition.
When the animation is this fast, I think it's ok with the "underline" effekt.

PR #227, opened 2024-11-26
DRA-1488 - Related videos not loading nicely.

jesperlauridsen:
This is only testable on devel, since kaltura still rejects the localhost to play videos. So build this branch to devel if you wanna test it (as I've done). Should fix the issue with the content not changing when pressing related videos.


Pbruun:
Looks good.

PR #228, opened 2024-11-26
DRA-1471 - cookie script path changed.

jesperlauridsen:
Hopefully pretty simple. This works on my machine, and hopefully does the same on devel/stage/prod when we get there too.


Pbruun:
Works for me as well. I'll merge

PR #230, opened 2024-11-26
DRA-1495: Autocomplete terms should have weights in the title

jesperlauridsen:
This actually fixes 3 things:

1: tabbing though did not change the search term before click - now fixed.
2: tabbing through changed the layout slightly on the elements - now fixed.
3: Added term weights (numbers) to the autocomplete - new feature.


Pbruun:
Worked great. Merging

PR #232, opened 2024-11-27
DRA-1486 not allowed text and color update

Pbruun:
Link needed to be visible.
Some text changes.


jesperlauridsen (approved):
Looks good.

PR #233, opened 2024-11-27
DRA-bugfix: error when searching after timesearch.

jesperlauridsen:
Was reported by UX - search after something, apply time filters and search again. Before, it did a search on the previous searchterm instead of the new one. Fixed now.


Pbruun:
Works as intended. Shouldn't the lines be deleted?

jesperlauridsen:
> Works as intended. Shouldn't the lines be deleted?

Fair point. Removed :-) Can merge now if happy with it!

Pbruun:
Merging! :)

PR #238, opened 2024-12-02
English translations.

Pbruun:
Added localization where needed. Made some custom error pop-ups.


jesperlauridsen (approved):
Very nice work, gjgj!

PR #244, opened 2024-12-03
Made 3 words cursive in search guide

Pbruun:


jesperlauridsen (approved):
Seems to work. We're not a fan of this per se, but we'll make an exception for this, just this once.

PR #248, opened 2024-12-04
Changed css class names

Pbruun:
The class names for categories had changed. Fixed with correct naming


jesperlauridsen (approved):
Well done!

PR #249, opened 2024-12-04
Categories translated in the application

jesperlauridsen:
categories now translated in the following places:

In main categories on the frontpage.
In the facets when you open and select.
In the facet headline when you select/deselect
In the full programme post when where you can click.


Pbruun:
Awesome! Great to have this

PR #250, opened 2024-12-04
Bugfix - category and genre selected in headline not getting loaded on a refresh on searchpage

jesperlauridsen:
This fixes it.


Pbruun:
Great find!

PR #251, opened 2024-12-04
DRA-1497 - tracking scripts added

jesperlauridsen:
This should add the tracking scripts that UX wants.


Pbruun:
Looks fine. Merging.

PR #256, opened 2024-12-05
better reset button

jesperlauridsen:
better reset button functionality. Now it's the same in both CurrentFilters and SearchOverhead. Good catch.


Pbruun:
Great work, I'll merge

PR #259, opened 2024-12-05
Updated all dependencies.

jesperlauridsen:
Updated all our dependencies, there were a few high risk vulnerabilities. They're all gone now. Try to spin up the app and see that everyhing works nicely.


Pbruun:
I see no issues

PR #262, opened 2024-12-09
Fewer facet calls to the backend

jesperlauridsen:
We found a place to reduce some backend calls - when the facet strings for the 2 backend facet calls are the same, we can do only one call - especially since we know this will happen anytime someone has done a new search and open the facets for the first time.


Pbruun:
Works great. Good work

PR #265, opened 2024-12-13
DRA-1532: removed the title and closed the categories container.

jesperlauridsen:
On the frontpage, the categorys are now closed, and can be folded out, in all the layouts (before, it was open at the desktop breakpoint and forward.)
On the nohits page, the categorys are still folded out on all layouts, and the description is added. (because it's nice that way)


Pbruun:
I approve these changes

PR #266, opened 2024-12-16
DRA-1529 - hide breadcrumb nicely

jesperlauridsen:
First stab - not finished


Pbruun:
Looks great. Only issue I found was that it collapses while hovering over DR-arkivet button and later breadcrumbs.

danielqvistcom:
Can only re-create the collapse-thing Patrick mentions. If I open breadcrumb, hover DR-Arkivet, and then back to breadcrumb once more. Don't think it's a "show-stopper"  :)    And I agree - looks great!

PR #267, opened 2024-12-19
Dra write page number

Pbruun:
On the search page - added an input for a page number.
Now scrolls to the top of window when changing a page
Pager now only displays 5 pages on less than 850px in width. 10 pages when above.

### DIFF
--- a/src/components/search/Pager.vue
+++ b/src/components/search/Pager.vue
@@ -49,12 +49,20 @@
 			</button>
 		</div>
 		<span class="pager-input">
+			<div class="input-label">{{ $t('search.page') }}</div>
 			<input
+				ref="inputRef"
+				v-model="selectPage"
 				:class="`page-select-input ${inputIncorrect ? 'page-select-input-error' : ''}`"
-				:placeholder="$t('search.pageNumberPlaceholder')"
 				@keydown="checkIfNumber($event)"
 				@keyup.enter="submitPage()"
 			/>
+			<button
+				:class="`go-to-button ${inputIncorrect || selectPage === '' ? 'go-to-button-error' : ''}`"
+				@click="submitPage()"
+			>
+				{{ $t('search.goToThePage') }}
+			</button>
 		</span>
 	</div>
 </template>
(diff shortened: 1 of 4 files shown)


jesperlauridsen:
image

You got some nice interactive checking going on here - but 2 points here:

When the number is invalid, the sorrunding color on the input box should be red, not orange.
When the number is invalid, we should also disable the button - right now it still activates the hover on the button, which it shouldn't, imho.
If you then press in the input field anyway (with fx enter), then we throw you the error.

src/components/search/Pager.vue
```diff
@@ -229,4 +302,80 @@ button span,
...
+
+.page-select-input-error {
+	box-shadow: 0 0 5px #f7ae3b;
+}
+
+.page-select-input-error:focus-visible {
+	outline: 1px solid #f7ae3b;
+}
+
+.go-to-button {
```
jesperlauridsen, line 349: This should have some sort of "transition: all 0.3s linear 0s;" or something, so it's a more smooth background color transition.

src/views/Search.vue
```diff
@@ -83,13 +83,15 @@ export default defineComponent({
...
 
-		const numPagesToShow = 8;
+		const numPagesToShow = ref(window.innerWidth < 850 ? 3 : 8);
 		const { t } = useI18n();
 
 		const errorManager = inject('errorManager') as ErrorManagerType;
-
 		onMounted(() => {
 			// we set the title of the archive here - needed if we go back from a page that sets it otherwise.
+			window.addEventListener('resize', () => {
```
jesperlauridsen, line 92: So, when you're adding an eventlistener, and adding a global function like this () =>  {}, it's stored in such away, that it's really hard to find it again, and just removing an eventlistener with the same global function, acutally removes nothing.

image


As you can see in this picture, when you go back to the frontpage and do a new search, it adds a new listener (from search line 92), but the old one isn't removed.

To fix this, the function you add must be a function called, and not a global function like () => {}

So it should be:

window.addEventListener('resize', updateResizeProperty);


And then put

numPagesToShow.value = window.innerWidth < 600 ? 3 : 8;

in a function.

jesperlauridsen: And then update the unmounted function accordingly ;-)

jesperlauridsen (changes requested):
A few pointers here too :-)

PR #268, opened 2024-12-19
DRA-1554: new headline text on filter searches.

jesperlauridsen:
The issue is described in DRA-1554. Very straight forward change.


Pbruun:
Looks good to me.

PR #269, opened 2025-01-06
Dra 1528 ask library

Pbruun:
Added the ask the library button on the needed areas.
Made it its own component for reuse.

### DIFF
--- a/src/components/search/ContactUs.vue
+++ b/src/components/search/ContactUs.vue
@@ -1,0 +1,97 @@
+<template>
+	<div :class="`contact ${relativePosition ? 'relative' : ''}`">
+		<h1>{{ t('search.contactUs') }}</h1>
+		<a
+			class="btn-blue btn"
+			:href="t('footer.column4.links.1.link')"
+			:data-testid="addTestDataEnrichment('button', 'ContactUs', 'link-to-ask-about-library', 0)"
+		>
+			<p>{{ t('footer.column4.links.1.title') }}</p>
+			<span class="material-icons">chat_bubble</span>
+		</a>
+	</div>
+</template>
+<script lang="ts">
+import { defineComponent, PropType } from 'vue';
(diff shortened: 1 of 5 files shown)


src/views/404.vue
```diff
@@ -41,26 +41,19 @@
 				</div>
 			</div>
 		</div>
-		<div class="contact">
-			<h1>{{ t('search.contactUs') }}</h1>
```
jesperlauridsen, line 45: I don't think this should be a <h1> element. It's never at the top of the page.

src/components/search/ContactUs.vue
```diff
@@ -0,0 +1,101 @@
+<template>
+	<div :class="`${relativePosition ? 'contact-relative' : 'contact'}`">
+		<h1>{{ t('search.contactUs') }}</h1>
```
jesperlauridsen, line 3: I don't think this should be a < h1 > element. It's never at the top of the page.

src/components/search/NoHits.vue
```diff
@@ -89,6 +89,16 @@
 					</div>
 				</template>
 			</EdgedContentArea>
+			<EdgedContentArea
```
jesperlauridsen, line 92: Why is this wrapped in a EdgedContentArea when we're not using the edges, backgroundcolor or anything here? It just adds a lot of unnecessary white space around the element

jesperlauridsen (changes requested):
I got a few pointers here :-)

src/components/search/ContactUs.vue
Pbruun: Good to know, made it into h2 instead

src/components/search/NoHits.vue
Pbruun: Replaced it with styling instead

jesperlauridsen (approved):
This looks good! Gjgj!

PR #270, opened 2025-01-13
DRA-1444: Handling no record ID on record page.

jesperlauridsen:
This should solve the problem of a manipulated record url. If you find a record and change it around to a record that doesn't exist, like http://localhost:3000/find-materiale/dr-arkivet/post/ds.tv:oai:io:fd596c49-0836-4546-b131-20ec25c6b11edsadsa - you'll now get a decent "we cant find this"-page instead of just an empty one.

### DIFF
--- a/src/components/common/MainCategories.vue
+++ b/src/components/common/MainCategories.vue
@@ -201,6 +201,7 @@ export default defineComponent({
 	justify-content: center;
 	min-height: 200px;
 	box-sizing: border-box;
+	padding: 0px 12px;
 }
 
 .category-item {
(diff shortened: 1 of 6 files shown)


Pbruun:
Looks great! The only thing I think looks a bit odd is the contact us button seems a bit floaty. Maybe something similar to no result page would fit? But that might just be me.

Pbruun:
@jesperlauridsen To me I would like the button to align with the left most buttons either above or below.

jesperlauridsen:
@Pbruun Are you happy now?! :D

Pbruun:
Now I'm happy! :D Good work, I'll merge!

PR #272, opened 2025-01-16
Bug fix search query

Pbruun:
We had an issue if you made a search. Then search the exact same thing with no filters or page changes. It would clear search result store. But because nothing changed it wouldn't update things in search.vue.  Made it so it does not clear or route.push if you're on the search page, and current query and last query are the same.

### DIFF
--- a/src/components/search/SearchBar.vue
+++ b/src/components/search/SearchBar.vue
@@ -340,28 +340,33 @@ export default defineComponent({
 
 		const search = () => {
 			if (selectedPortal.value === 'drArchive') {
-				searchResultStore.searchResult = [];
-				searchResultStore.resetAutocomplete();
-				searchResultStore.channelFilters = [];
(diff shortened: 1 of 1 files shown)


Pbruun:
Spotted an issue with my solution. Will not reset filters on the same search.

src/components/search/SearchBar.vue
```diff
@@ -340,7 +340,14 @@ export default defineComponent({
 
 		const search = () => {
 			if (selectedPortal.value === 'drArchive') {
-				searchResultStore.searchResult = [];
+				console.log(searchResultStore.lastSearchQuery);
```
jesperlauridsen, line 343: Cant have console.logs here, brother! ;)

Pbruun: Fixed! :D

PR #273, opened 2025-01-16
DRA-1531: This should fix the issues with the column layout positioning.

jesperlauridsen:
This should address the issues in 1531:

- Now the items in the channels flow down, then to the next column, as requested by UX.
- Added a hover effect that shows when you hover over them which checkbox is associated with which title - if the borders aren't enough.


danielqvistcom (commented):
Looks great, and it helps out with the hover.
If it doesn't cause to much trouble, then we could float the checkmarks - to the left?

PR #274, opened 2025-01-22
DRA-1566 dropdown close on click

Pbruun:
Added eventlistener on click events, checks if they're part of the dropdown element, if not, then close the dropdown.

### DIFF
--- a/src/components/search/SearchBar.vue
+++ b/src/components/search/SearchBar.vue
@@ -261,18 +261,19 @@ export default defineComponent({
 		const closePortal = (e: MouseEvent) => {
 			if (selectButtonRef.value instanceof HTMLButtonElement && e.target instanceof Node) {
 				if (!selectButtonRef.value.contains(e.target)) {
-					showPortalSelector.value = false;
+					togglePortalSelector();
 				}
 			}
 		};
 		onMounted(() => {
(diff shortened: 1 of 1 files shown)


src/components/search/SearchBar.vue
```diff
@@ -258,13 +258,21 @@ export default defineComponent({
...
+		const closePortal = (e: MouseEvent) => {
+			if (selectButtonRef.value instanceof HTMLButtonElement && e.target instanceof Node) {
+				if (!selectButtonRef.value.contains(e.target)) {
+					showPortalSelector.value = false;
+				}
+			}
+		};
 		onMounted(() => {
 			window.addEventListener('toggle-search', toggleSearchField);
+			document.addEventListener('click', closePortal);
```
jesperlauridsen, line 270: We're almost there. I think this is really nice, but I would like us to only add the event listener when the portal selector is opened up - and remove it if it's closed or on unMounted. That way, we make sure it's only active whenever we actually need it.

jesperlauridsen (changes requested):
Just a single observation :-)

src/components/search/SearchBar.vue
Pbruun: Should be fixed now

jesperlauridsen (approved):
Looks good!

PR #276, opened 2025-01-23
DRA-1478 autocomplete pick correct element

Pbruun:
If you hover over an autocomplete element. Then use arrows or tab to select other element. 
Before if you then clicked with your mouse, it would choose the selected element from the arrows. 

This fix should make sure, if you mouse click an element, it will always choose the clicked element. 
With enter, it should pick the active element


jesperlauridsen (approved):
Very nicely done!

PR #277, opened 2025-01-23
DRA-BUGFIX imageData no longer throw warning

Pbruun:
ImageData threw warnings because it was sat as type: Object as PropType<string>, but the data it got was String.
My IDE displays  as if it had an error. Had to change the method in style, so it wouldn't report falsely


jesperlauridsen (approved):
Pretty straightforward \o/

PR #278, opened 2025-01-27
Added alternate-reverse to animation

Pbruun:
This should fix it. Seems to work on mac now. And I don't see any issues on other browsers


jesperlauridsen (approved):
Awesome find man, gj!

PR #279, opened 2025-01-29
DRA-769-Copy video timestamp

Pbruun:
Added a button where you can copy the link for where the video currently is at.
It checks the timestamp every 1.5 seconds. And will only update if the difference is 3 seconds.

### DIFF
--- a/src/components/records/BroadcastAudioRecord.vue
+++ b/src/components/records/BroadcastAudioRecord.vue
@@ -89,6 +89,23 @@
 						<a class="link">{{ $t('record.copy') }}</a>
 					</div>
 				</div>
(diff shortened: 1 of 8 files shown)


Pbruun:
Need some clarification whether or not we can have this feature. Also need UX design. So do not merge to main.

kesj-kb:
We are not allowed to merge this PR.
From Louise Ahrenfeldt: "En fremsøgt udsendelse skal altid starte fra begyndelsen".

jesperlauridsen:
This is now, unfortunately, officially rejected. According to the agreements we've signed, we cannot do any kind of deep linking.

PR #282, opened 2025-02-21
Added testid

Pbruun:
In order to test if the through time segment is visible.


jesperlauridsen (approved):
I don't see any issues with this. Approved.

PR #284, opened 2025-02-25
Added some test-ids to the portal dropdown

jesperlauridsen:
Very simple, just need some test-ids here :-)


Pbruun:
I do not see an issue with this :)

PR #285, opened 2025-02-27
Dra 1682 limit last page

Pbruun:
We now make sure to limit max pages.
A message box appears if at the limit, and there's more to be seen.

Need review on the styling.

### DIFF
--- a/src/components/search/Pager.vue
+++ b/src/components/search/Pager.vue
@@ -1,5 +1,24 @@
 <template>
 	<div :class="searchResultStore.loading ? 'pager disabled' : 'pager'">
+		<div class="morePagesMessage">
+			<div v-if="totalPages > maxPages && currentPageRef === maxPages">
+				<div>
+					{{ $t('search.morePagesMessage.1') }}
+					{{ $t('search.morePagesMessage.2') }}
+					<div class="totalHits">
+						{{ $t('search.morePagesMessage.3', { totalHits: totalHits.toLocaleString('da-dk') }) }}
+					</div>
+					{{ $t('search.morePagesMessage.4') }}
+					<a
+						href="#"
+						class="topOfScreen"
+					>
+						<span class="material-icons">sort</span>
+						{{ $t('search.morePagesMessage.5') }}
+					</a>
+				</div>
+			</div>
+		</div>
 		<div class="pager-buttons">
 			<button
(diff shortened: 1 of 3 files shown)


danielqvistcom:
Looks great! Ill get back with text-updates asap :)

danielqvistcom (approved):

src/components/search/Pager.vue
```diff
@@ -374,6 +395,28 @@ button span,
...
+}
+.morePagesMessage .totalHits {
+	font-weight: bold;
+	width: fit-content;
+	display: inline;
+}
+.morePagesMessage .topOfScreen {
+	color: #002e70;
+}
+.topOfScreen > span {
```
jesperlauridsen, line 416: image

This floats!
doing:

    position: relative;
    top: 7px;
    line-height: 9px;
    
instead fixes the weird gap between the last and the second to last line.

jesperlauridsen (changes requested):
This looks very well - only one minor comment to fix, imho :-)

Edit: I take that back - see next comment :D

jesperlauridsen:
When you switch to grid, we show 40 entrys per page, and then the limit should be 25  and not 100. Because if you go over, we get an error from the backend.

src/components/search/Pager.vue
```diff
@@ -1,5 +1,24 @@
...
+		<div class="morePagesMessage">
+			<div v-if="totalPages > maxPages && currentPageRef === maxPages">
+				<div>
+					{{ $t('search.morePagesMessage.1') }}
+					{{ $t('search.morePagesMessage.2') }}
+					<div class="totalHits">
+						{{ $t('search.morePagesMessage.3', { totalHits: totalHits.toLocaleString('da-dk') }) }}
+					</div>
+					{{ $t('search.morePagesMessage.4') }}
+					<a
```
jesperlauridsen, line 12: Having a link here is okey - I prefer the solution we've done other places with a scroll to the top instead, with a javascript window.scrollTo. This gives a smoother experience instead of the direct jump to the top.

jesperlauridsen:
image
Right now, if you hover over one of the arrows < / > and you're on the last or first page, so there no more pages to go, it's gray, but the hover still shows a a click cursor. That should probably be a normal cursor instead.

PR #286, opened 2025-03-03
Time difference with days/ weeks/ months/ years

Pbruun:
Instead of just showing approx year. Even if the filter is only 1 day.
We now split it between days, weeks, months and years.


src/locales/en.json
```diff
@@ -48,7 +48,9 @@
       "searchCategories": "Search by category",
       "year":"year | years",
       "month":"month | months",
+      "week":"week | weeks",
       "day":"day | days",
+      "weekday":"day | days",
```
jesperlauridsen, line 53: Seeing as day and weekday are the same here - could we not just reuse the day translation here?

src/components/search/CurrentFilters.vue
```diff
@@ -229,6 +224,25 @@ export default defineComponent({
 			});
 		};
 
+		const approxTimeDifference = () => {
+			const yearDifference: number = endDate.value.getFullYear() - startDate.value.getFullYear();
+			const monthDifference: number = endDate.value.getMonth() - startDate.value.getMonth();
+			const dayDifference: number = endDate.value.getDate() - startDate.value.getDate();
+			if (yearDifference === 0) {
+				if (monthDifference === 0) {
+					if (dayDifference >= 7) {
```
jesperlauridsen, line 233: Imho, 7 days equals a week, so it should be 6~ days, and the next step up would be "~1 week", and not 7 days.I don't have a strong opinion on this, so I'll let you decide - it was just something i noticed.

jesperlauridsen:
@Pbruun can you resolve merge conflicts here, pretty please :D

jesperlauridsen (approved):
This checks out now - good work! \o/

PR #287, opened 2025-03-11
Dra 1730 rewrite header menu

Pbruun:
We wish to go away from webcomponents and instead use Vue as the rest of the project. 
Rewrote the webcomponents as Vue, keeping the style and functionality.


src/components/global/nav/HeaderMenu.vue
```diff
@@ -0,0 +1,866 @@
...
+											{{ item.icon }}
+										</i>
+									</button>
+									<a
+										v-else
+										:href="item.link"
+										:data-testid="addTestDataEnrichment('link', 'topmenu', item.title, 0)"
+										class="nav-item level-1"
+									>
+										{{ item.title }}
```
jesperlauridsen, line 173: Minor, MINOR detail. But when title is inserted like this, we get an extra space, because of the way the templates work with our syntax. If you inspect on devel versus localhost on the same screensize, you'll notice everything jumps a little to the left. doing 

<span v-text="item.title"></span>

instead will remove the single space after "login"

src/components/global/nav/HeaderMenu.vue
```diff
@@ -0,0 +1,866 @@
...
+		const { t, messages, locale } = useI18n();
+		const currentLocaleMessages = ref(undefined as unknown as HeaderType);
+		let currentLocale = ref('da');
+		const mainHeaderRef = ref<HTMLFormElement | null>(null);
+		const menuOpen = ref(false);
+		const router = useRouter();
+		const route = useRoute();
+
+		const switchLocale = (e: Event) => {
+			e.preventDefault();
```
jesperlauridsen, line 218: Why the preventDefault() here?

src/components/global/nav/HeaderMenu.vue
```diff
@@ -0,0 +1,866 @@
...
+	name: 'HeaderMenu',
+	props: {
+		searchBarOpen: { type: Boolean as PropType<boolean>, required: true },
+	},
+
+	emits: ['toggleSearchBar'],
+	setup(props, { emit }) {
+		const { t, messages, locale } = useI18n();
+		const currentLocaleMessages = ref(undefined as unknown as HeaderType);
+		let currentLocale = ref('da');
```
jesperlauridsen, line 211: Is this currentLocale variable needed? Can't we just look at the locale from the i18n?

src/components/global/nav/HeaderMenu.vue
```diff
@@ -0,0 +1,866 @@
...
+.search-button {
+	display: flex;
+	flex-direction: column;
+	align-content: center;
+	align-items: center;
+}
+
+.rdl-logo {
+	/* TODO: We need a better way to get images into web components. */
+	background-image: url('https://design.kb.dk/components/assets/images/logo-digital.svg');
```
jesperlauridsen, line 299: We should get this into our repo instead of here now. This was some legacy stuff we thought would work in the long run.

src/locales/da.json
```diff
@@ -17,6 +17,61 @@
...
+            },{
+               "title": "Services", 
+               "link": "https://www.kb.dk/services"
+            },{
+               "title": "Besøg os", 
+               "link": "https://www.kb.dk/besoeg-os"
+            },{
+               "title": "Luk", 
+               "altTitle": "Søg",
+               "link": "#",
```
jesperlauridsen, line 69: Does this need to here at all, when it's a button?

src/types/HeaderType.ts
```diff
@@ -0,0 +1,11 @@
+export interface HeaderType {
+	[header: string]: [
+		{
+			title: string;
+			altTitle: string;
+			link: string;
+			id: string;
```
jesperlauridsen, line 7: I'm not sure, but wouldn't it be the best practice to add a ? after optional attributes - like id, altTitle, link etc?

Pbruun: Good point, I think that makes great sense

src/locales/da.json
Pbruun: It doesn't, so I have removed it

src/components/global/nav/HeaderMenu.vue
Pbruun: Good question, works without, so I have removed it.

src/components/global/nav/HeaderMenu.vue
Pbruun: Doesn't seem like it

src/components/global/nav/HeaderMenu.vue
Pbruun: Can't see a whitespace. But found we're setting margin-left: 4px on it. So I have removed that

src/components/global/nav/HeaderMenu.vue
Pbruun: We already had it in our repo. Changed url to go to our assets folder instead

src/components/global/nav/HeaderMenu.vue
jesperlauridsen: image
VS
image


The CSS looks exactly the same, I'm quite sure its the space made from the templates (which is a known issue because of the linebreaks syntax), causes this.

jesperlauridsen: This is some serious nitpicking, I know, but the margin-left should be re-added and the span with v-text added instead. I'll showcase on monday why. But it's some _SERIOUS_ nitpicking, sorry!

jesperlauridsen (approved):
Looks good now.

PR #288, opened 2025-03-17
Dra 1731 header from drupal

Pbruun:
We now use the api to give data to the header. 
Made a store to handle the api data. 
Made the initial steps for footer data. But we're not to implement new footer yet.


jesperlauridsen:
@Pbruun could you resolve the merge conflicts here?

PR #289, opened 2025-03-19
Display total results and sort icon

Pbruun:
We wish to display, how many results are in the current page, and what that amounts to out of the total result count.

Added pageNumber to store, because I wished to reuse it, and it just became convoluted math in the html.

### DIFF
--- a/src/components/search/Sort.vue
+++ b/src/components/search/Sort.vue
@@ -1,5 +1,6 @@
 <template>
 	<div class="sort">
+		<span class="material-icons sort-icon">sort</span>
 		<p class="sort-by">{{ t('search.sortBy') }}:</p>
 		<button
 			ref="relevanceRef"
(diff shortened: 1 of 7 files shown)


jesperlauridsen:
image
Should the second number not be formatted like the top one?

src/components/search/Sort.vue
```diff
@@ -148,7 +149,7 @@ export default defineComponent({
 	box-sizing: border-box;
 }
 
-.sort .material-icons {
+.sort > button > .material-icons {
```
jesperlauridsen, line 152: Is there a specific reason why you've chosen this kinda selector?

Pbruun: Because we got another .sort .material-icons. The new sort icon, which will be invisible otherwise

jesperlauridsen:
@danielqvistcom 

image


this seems very bloated, because they numbers are almost exactly alike. Could it not be something like "resultatsæt 1-10" or something, without the of XXXXX?

danielqvistcom:
I think it's a couple of good points. The format's should align. xx.xxxx
And maybe bigger difference - with tekst update like "resultatsæt". I'll talk to UX content later today, and get back with update ASAP.

PR #290, opened 2025-03-20
A lot of small accessablity fixes.

jesperlauridsen:


src/components/global/content-elements/EdgedContentArea.vue
```diff
@@ -261,8 +262,8 @@ h3:first-letter {
 /* MEDIA QUERY 1280 */
 @media (min-width: 1280px) {
 	.time-search .container {
-		padding-right: 0;
-		padding-left: 0;
+		/* padding-right: 0;
+		padding-left: 0; */
```
Pbruun, line 266: Code commented out

src/components/search/ResultItem.vue
```diff
@@ -18,15 +19,34 @@
 							:title="resultdata.title"
 						>
 							{{ resultdata.title[0] }}
-							<span class="material-icons arrow">keyboard_arrow_right</span>
+							<div
+								role="img"
+								class="material-icons arrow"
+								:aria-label="`${t('app.a11y.goToPost')}`"
```
Pbruun, line 25: I think it would be cleaner with :aria-label="t('app.a11y.goToPost')"

src/components/search/ResultItem.vue
```diff
@@ -18,15 +19,34 @@
...
 						<div class="subtitle">
-							<span class="material-icons icons schedule">
-								{{ resultdata.origin.split('.')[1] === 'tv' ? 'play_circle_filled' : 'volume_up' }}
+							<span
+								role="img"
+								:class="`icons schedule material-icons ${
+									resultdata.origin.split('.')[1] === 'tv' ? 'playSVG' : 'volumeSVG'
+								}`"
+								:aria-label="`${t('app.a11y.broadcastTimeAndPlace')}`"
+							>
```
Pbruun, line 37: Same as above

src/components/search/ResultItem.vue
```diff
@@ -18,15 +19,34 @@
...
+								play_circle_filled
 							</span>
 							<span class="where">{{ resultdata.creator_affiliation + ',' }}</span>
 							<span class="when">{{ starttime }}</span>
-							<span class="material-icons icons schedule">schedule</span>
+							<div
+								role="img"
+								class="material-icons icons schedule timeSVG"
+								:aria-label="`${t('app.a11y.broadcastDuration')}`"
```
Pbruun, line 45: Same as above

src/components/search/ResultItem.vue
```diff
@@ -61,23 +81,31 @@
...
 					<div
 						ref="placeholderTitleRef"
 						class="placeholder-t"
 						:style="`width:${Math.random() * 20 + 20 + '%'}`"
 					></div>
 					<div class="placeholder-w">
-						<span class="material-icons ph-icon">play_circle_filled</span>
+						<div
+							role="img"
+							:aria-label="`${t('app.a11y.broadcastTimeAndPlace')}`"
```
Pbruun, line 93: Same as above

src/components/search/ResultItem.vue
```diff
@@ -61,23 +81,31 @@
...
+						</div>
 						<span
 							class="line"
 							:style="`width:${Math.random() * 10 + 10 + '%'}`"
 						></span>
-						<span class="material-icons ph-icon">schedule</span>
+						<div
+							role="img"
+							class="material-icons ph-icon"
+							:aria-label="`${t('app.a11y.broadcastDuration')}`"
```
Pbruun, line 105: Same as above

src/components/search/SoundThumbnail.vue
```diff
@@ -7,23 +7,33 @@
 				:alt="`Image of soundwaves for ${resultTitle}`"
 			/>
 		</div>
-		<span class="material-icons play">play_circle_filled</span>
+		<span
+			role="img"
+			:aria-label="`${t('app.a11y.soundIcon')}`"
```
Pbruun, line 12: :aria-label="t('app.a11y.soundIcon')"

Pbruun:
Screenshot from 2025-03-23 15-10-58
The icon in the button does not align anymore

Pbruun:
image
Going to the last page on the pager, has issues with the text.

Pbruun:
On the search page. If you switch between sorting. It now reloads the data, while also making the elements move slightly.

jesperlauridsen:
> On the search page. If you switch between sorting. It now reloads the data, while also making the elements move slightly.

it has always reloaded the data - which is because we only get 10 results of the set (the 10 first), but sorting needs to be done on the entire set (which could be thousand or more) so we gotta reload with the new sort to get the real first of the new sort from the backend.

src/components/search/ResultItem.vue
jesperlauridsen: Fill fix, good idea.

PR #292, opened 2025-04-01
Audit maintenance

jesperlauridsen:
We had a few vulnerabilites, that is sorted now. Please test that

You can still run npm run dev
That you can still build to devel
That core functionality works as expected (which is same as before, really)


Pbruun:
Seems to work as expected

PR #298, opened 2025-04-15
DRA-BUGFIX: 2 small bugfixes

jesperlauridsen:
first one makes sure we reset start=0, when we toggle tv/radio - that way we never get out of bounce.

second one was a simple change to the dates match if they're at the end of the spectrum - this way the filter is only added if it is actually set.


Pbruun:
Tested it, seems to work as intended :smiley: I'll merge

PR #299, opened 2025-04-15
DRA-1484: new title setup

jesperlauridsen:
So, this is a rework of how we set the title at the Search page. There is now a function, looking at what params the search was based on. If there was a q, we use that. If not, we look at the different filters and find out exactly how we want the title to look.

Have a look through and test it, and see if it holds up.

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -10,7 +10,17 @@
             },
             "about": "About"
          },
-         "search":" Search for"
+         "search":" Search for",
+         "SearchWith":"Søgning with",
+         "PeriodSearch":"Søgning in the period",
+         "to":"to",
+         "filter":"filter | filters",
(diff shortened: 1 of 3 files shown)


src/locales/en.json
```diff
@@ -10,7 +10,17 @@
             },
             "about": "About"
          },
-         "search":" Search for"
+         "search":" Search for",
+         "SearchWith":"Søgning with",
```
Pbruun, line 14: Shouldn't this be "Searching"?

src/locales/en.json
```diff
@@ -10,7 +10,17 @@
...
+         "PeriodSearch":"Søgning in the period",
```
Pbruun, line 15: Same as above

Pbruun:
Didn't find anything else, I'll merge :)

PR #300, opened 2025-05-01
Bugfix: disable picture-in-picture for firefox builtin functionality

jesperlauridsen:
Fixed so we always remove the picture in picture possiblity on all browsers (firefox has a builtin that showed up regardless of kalturas disabled setting)

Also fixed a small mis-alignement where an error message was missing in the audioplayer.


Pbruun:
Tested on chrome and firefox, seems to work as intended :smiley: I'll merge

PR #301, opened 2025-05-01
DRA-1910: Change pager buttons to links

jesperlauridsen:
We changed the buttons to links instead - Monsido wants it this way.
Also removed some useless metadata that was kinda wrong for how our site is build. Monsido recommendation aswell.


Pbruun:
Works great, I see no issues with this

PR #302, opened 2025-05-09
DRA-1882: A better /find page with no params set.

jesperlauridsen:
This PR is doing 2 things:

1: Making sure any reset of filters on *:* searches now also removes the query, sending you to an empty /find page.
2: Restyled the /find page after the UX-design provided by Daniel - which can be found in the JIRA ticket.


Pbruun (approved):
I see no issues. Approved!

PR #303, opened 2025-05-16
DRA-1665: fixed some styling bugs

jesperlauridsen:


danielqvistcom (commented):
Only minor ting:

image

Categories doesn't allign with the margin of the page...
And the search-icon is still yellow? :)

PR #305, opened 2025-05-22
DRA-1974: New audio player

jesperlauridsen:
Added styling and destruction of players to make sure audio and video uses the right player when loading.

Important: If this is added, any new release must ensure that the BFF service messages with id's for the player is updated accordingly, so everything matches.


danielqvistcom (commented):
Player = nice!
Channel-logo approval = Pending!...  :(

Pbruun (changes requested):
Just a few cleanup things, otherwise it looks great!

PR #307, opened 2025-06-16
DRA-1998 font setting changes

Pbruun:
Made asked for changes to follow font settings from kb.dk
because with the line height changes to grid display. I have made the size of the card 15 px larger. 
Icons on the programme display had 10px margin to the right on kb.dk. But because we use 7px gap already, I added a 3px margin instead.


danielqvistcom (approved):
Nice, det ser fint ud. (OBS -> Jeg kan ikke teste på mobil)

jesperlauridsen (approved):
Looks fine

PR #311, opened 2025-06-26
Dra 2012 add episode number

Pbruun:
Added episode number for record display and search records. 
It only display if a record has it in the metadata.


jesperlauridsen (approved):
This looks very well!

PR #312, opened 2025-06-27
DRA-OPTIONAL channel selection cleaner design

Pbruun:
Selected channels now create an overflow button.
Overflow button displays the amount of selected channels, that are not shown in the overview.
Eventlistener on the size of the window, to make sure the design is dynamic.


jesperlauridsen (approved):
This looks very nice!

PR #316, opened 2025-08-08
DRA-1809: Date picker refinement.

jesperlauridsen:
This should ensure that if you enter 1-1-16 (fx), it is interpreted as 1-1-2016, and thus becomes valid (the calendar component actually does that right now, our own logic did not. This should align the two)

### DIFF
--- a/src/utils/datepicker-utils.ts
+++ b/src/utils/datepicker-utils.ts
@@ -16,12 +16,14 @@ const updateSelectedDate = (
 	dateInput = dateInput.replace(/[.-]/g, '/');
 	const splitDateInput = dateInput.split('/');
 	if (splitDateInput.length === 3) {
-		const [day, month, year] = splitDateInput.map(Number);
+		const [day, month, yearInitial] = splitDateInput.map(Number);
+		let year = yearInitial;
 		// Validate if day, month, and year are numbers
 		if (!isNaN(day) && !isNaN(month) && !isNaN(year)) {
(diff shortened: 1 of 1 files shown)


src/utils/datepicker-utils.ts
```diff
@@ -16,12 +16,13 @@ const updateSelectedDate = (
 	dateInput = dateInput.replace(/[.-]/g, '/');
 	const splitDateInput = dateInput.split('/');
 	if (splitDateInput.length === 3) {
-		const [day, month, year] = splitDateInput.map(Number);
+		const [day, month, yearInitial] = splitDateInput.map(Number);
+		let year = yearInitial;
 		// Validate if day, month, and year are numbers
 		if (!isNaN(day) && !isNaN(month) && !isNaN(year)) {
 			// Months in JavaScript Date are 0-indexed, so subtract 1 from the month
+			if (year.toString().length === 2) year = Number(`20${year}`);
```
Pbruun, line 24: Just a thought to when we get older data. 
If the 2 digit number is beyond current year, then indead of  `20${year}` it should be `19${year}`.

Pbruun:
Seems to work as intended. I'll approve :)
Only thing with the input, is if you have selected a leap year, then change the year. It'll give a incorrect date, which is correct, but maybe a more userfriendly input, would be to change to the 28 of february?

Pbruun (approved):

PR #320, opened 2025-08-21
DRA-2003 need distance between the borders of the browser

Pbruun:
I only saw that we had an issue with this with the footer. Followed the padding used on kb.dk


jesperlauridsen (approved):
OK!

PR #322, opened 2025-08-21
DRA-1981 noHits search by categories

Pbruun:
Search by categories did not have the same style on noHits as the frontpage, when using mobile. So I have removed the padding which seemed to have fixed it.


jesperlauridsen (approved):
Fine!

PR #324, opened 2025-08-25
Added support for backend severity variable

jesperlauridsen:
Now, if severity is added to the reponse, we take it into acount. If it's 1, we use the green success, 2 for the white info, 3 for the orange warning and 4 nothing added, red error.

Defensively coded so if nothing is added, its as it used to be.


Pbruun:
Looks good :smiley: I approve!

PR #326, opened 2025-08-25
Grid item fix sizing between items with or without episodes

Pbruun:
Added missing padding. Added hidden div, to keep size


jesperlauridsen (approved):
Yay

PR #332, opened 2025-12-05
DRA-2274: New accessability-improved date picker.

jesperlauridsen:
--- THIS IS A DRAFT ---

This is like 90% of the picker done. Im still figuring our some minor isseus here and there, but the main part should be fairly done so you can review it, codewise.

Also, its only on the DayPicker right now (which is the one on the frontpage), the 2 pickers in the filters have not yet been updated

### DIFF
--- a/src/components/common/TimePicker.vue
+++ b/src/components/common/TimePicker.vue
@@ -8,7 +8,7 @@
 			:for="inputId"
 			class="visually-hidden"
 		>
-			Select date
+			{{ t('calendar.select') }}
 		</label>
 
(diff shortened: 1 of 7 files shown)


Pbruun:
No longer any issues with the calendar, great work!
The calendar button, does not have a pointer cursor, same with the month/ year dropdown and the arrows for next/ previous month.

A very nitpicky comment, the next/ previous month arrows seem very "default". Maybe something like the timemachine arrows? Up to you what you think fits.

Can't find or think of anything else. :smiley:

PR #350, opened 2026-04-10
Annotation for a program added

Pbruun:
Annotation added for programs. 
The field is single valued, so we do not get an english text.

### DIFF
--- a/src/types/BroadcastRecordType.ts
+++ b/src/types/BroadcastRecordType.ts
@@ -6,6 +6,7 @@ export interface BroadcastRecordType {
 	id: string;
 	name: NameType[];
 	publication: PublicationType;
+	annotation: string;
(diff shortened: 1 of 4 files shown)


src/types/BroadcastRecordType.ts
```diff
@@ -6,6 +6,7 @@ export interface BroadcastRecordType {
 	id: string;
 	name: NameType[];
 	publication: PublicationType;
+	annotation: string;
```
jesperlauridsen: Is annotation always here? If it is and just empty sometimes, then cool. Otherwise this should be annotation?: string;, like encodesCreativeWork below it.

Pbruun: Great find. No the annotation will not alway be there (mostly not).

PR #351, opened 2026-04-20
Dra 2522 kb button design

Pbruun:
We want to create a unified button design.
This PR creates a reuseable component containing the basic button design and layout. 
We only want to give access to a icon to the left or right. 
The buttons we change from button/ router-link are the CTA buttons.


jesperlauridsen (approved):
I think this looks solid.

PR #352, opened 2026-04-20
Dra 2637 primary button padding

Pbruun:
This PR extends the KBButton component. Also making it useable with anchor tags. And creates the design for the primary buttons used thoughout


jesperlauridsen (approved):
Looks good.

PR #354, opened 2026-04-21
DRA-optional: reverts the changes so we can release soonish.

jesperlauridsen:
reverts the thumbnail changes back to what is currently on the main ds-image backend branch.


Pbruun:
We have following inside the BroadcastAudioRecord and inside the record-utils.ts:
props.recordData.identifier.find((obj) => obj.PropertyID === 'KalturaID') !== undefined;
Should this also be changed to fileid?

Pbruun:
My previous comment is not relevant, so it looks good, I'll merge :)

PR #357, opened 2026-04-24
Dra 2581 taps button design

Pbruun:
Tags for displaying selected filters on search page has now been implemented as a KBButton.
Fixed a bug where a single checkbox did not display (removed display: block from label)
Removed a redundant remove filter button


jesperlauridsen (approved):
Looks very solid.

PR #358, opened 2026-04-27
Dra 2586 sort design

Pbruun:
Implementation of a new sort design. 
Have reused the base idea around the KBButton, and implemented a KBButtonSort


jesperlauridsen (approved):
I think this looks good.

PR #361, opened 2026-04-27
DRA 2673 load behaviour on search page

Pbruun:
Checkbox skeleton load. Search result header skeleton. Result items skeleton load. Minor height changes

Previously the page "jumped" when changing filters. It should be more smooth now


jesperlauridsen (approved):
This looks very nice!

PR #364, opened 2026-04-29
Dra 2583 archive material dropdown

Pbruun:
Changed timeSearchFilter's dropdown according to new dropdown design. Dropdown content goes below graph.


jesperlauridsen (approved):
Good work!

PR #368, opened 2026-04-30
DRA BUGFIX title for tags

Pbruun:
Without a title, the hover showed the underlying title, which gave the wrong impression of functionality.


jesperlauridsen (approved):
Nice work!

PR #369, opened 2026-05-08
DRA OPTIONAL sort random

Pbruun:
We made a change to randomly sort time search component.
The change did not include the initial state of the component.
This should do the job.


jesperlauridsen:
I think this looks very nice. Any reason it's a draft, @Pbruun?

jesperlauridsen (approved):
Should do the trick, imho?

PR #370, opened 2026-05-11
Dra 2522 button consistency

Pbruun:
We still needed to implement KBButton throughout our buttons. This PR creates the needed stylings for the remaining buttons. 
We adjusted the line height according to a standard where buttons should have a line height of 1.


jesperlauridsen (approved):
I like it!

PR #371, opened 2026-05-18
DRA bugfix annotation on radio

Pbruun:
We missed adding annotation for radio. This will fix that.


Pbruun:
Added margin, because if description is long, it'll get very close to programguide.

jesperlauridsen (approved):
Looking good!

PR #378, opened 2026-06-04
DRA:2769: Thumbnail requests now use kalturaId instead of fileId.

jesperlauridsen:
Should make the fix  - so we have thumbnails asking with kalturaId instead of fileId.


Pbruun (approved):
Looks good! Works on devel for each page we use thumbnails.

PR #379, opened 2026-06-08
DRA OPTIONAL timeFacetsOpen

Pbruun:
We have a maybe unintended behavior when making a search via period. Because timeFacetsOpen remains false when the user clicks "Show xx.xxx results" from the frontpage. Then on the searchpage if you click show filters, and doesn't set date filter to true. The time period filter will be reset if another filter is added.


jesperlauridsen (approved):
Good catch!

PR #381, opened 2026-06-10
DRA OPTIONAL split active filter correctly

Pbruun:
There were no seperation line between the types of filters. This should also fix any issues with smaller screens


jesperlauridsen (approved):
Good cleanup. I think this looks nice. Didn't spot any errors, and it tests fine. GJ!

PR #382, opened 2026-06-16
Dra 2702 slider design update

Pbruun:
Redesign of the time search slider. Make it responsive so its usable on mobile and above. Same design funcitonality inside the new filter.


src/components/common/FilterDateExpander.vue
```diff
@@ -0,0 +1,321 @@
+<template>
```
jesperlauridsen: This doesn't seem to be in use anywhere (thank god!) - can it just be removed then? :O

jesperlauridsen (approved):
I removed an unused component. Otherwise it looks good and i'll merge!

PR #384, opened 2026-06-22
DRA BUGFIX year range incorrect

Pbruun:
if you made a period filter. then refresh. The year range is set to the init of 2006. Changed to computed, so the values follow if the init values are updated. Added a watch for that value, to update the slider data.


jesperlauridsen (approved):
Looks good, good find.

PR #385, opened 2026-06-22
DRA BUGFIX FacetsContainer ref is null

Pbruun:
If you click show filters quickly after refresh. The backdrop appears, but the filter does not. It was because the gsap cannot see the facetsContainer yet, because its still null.


jesperlauridsen (approved):
I think this works - without any bugs, but i find it a little funny. However, I cant really come up with anything better right now, so here we go!

PR #386, opened 2026-06-22
DRA BUGFIX unable to remove all filters from category

Pbruun:
When checking all channels. The button to remove those selected channels change to a label, so you cant remove them all. This makes sure the KBButton stays even if all is checked.


jesperlauridsen (approved):
Good catch!

PR #388, opened 2026-06-22
label fixes

jesperlauridsen:
Should fix the small issue with some missing label fonts


Pbruun:
Looks good. We still need to look at the rest of the filterscontainer. But I'll merge this :+1:

PR #390, opened 2026-06-22
DRA-BUGFIX: gray icons when disabled.

jesperlauridsen:
Right color on the icons we use (the svg ones) when disabled, we used a filter. That wasn't the right color, so now we use the gray svgs when disabled instead.

### DIFF
--- a/src/components/common/SimpleCheckbox.vue
+++ b/src/components/common/SimpleCheckbox.vue
@@ -203,6 +203,10 @@ export default defineComponent({
 	background-image: url('@/assets/icons/blue/diverse-blue.svg');
 }
 
+.display-image.svg.disabled {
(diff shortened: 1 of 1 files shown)


Pbruun (changes requested):
The diverse image does not display on the disabled icon. This is because the 2 checkboxes that uses diverse svg, does not have the diverse class. It uses .display-image.svg.

jesperlauridsen:
> The diverse image does not display on the disabled icon. This is because the 2 checkboxes that uses diverse svg, does not have the diverse class. It uses .display-image.svg.

Good catch. Fixed :-)

Pbruun (approved):
It works! I'll merge

PR #391, opened 2026-06-23
DRA-BUGFIX: Star searches did a weird thing when reloading the page.

jesperlauridsen:
We get a small flash of *:* before it's removed on reload, but i think we can live with that, since the functionality doesn't break now.


Pbruun:
Great find! As far as I can see, it works great. I'll merge

Pbruun (approved):

PR #392, opened 2026-06-24
Dra 2498 initial changes

Pbruun:
This is the groundwork with removing components and adding the two main colors througout the website.
Outcommented the searchbar and breadcrumb, so they can be readded in order to test navigation. They'll be removed at a different stage


jesperlauridsen (approved):
This is really good work! Awesome job!

PR #393, opened 2026-06-24
Dra 2757 header redesign

Pbruun:
We wish to simplify the header. 
Removed every link other than login and change language button. 
Had some issues with margin-top from content below header on mobile view. Maybe we don't need it any longer? Or it has to be added when hero is reimplemented.


jesperlauridsen (approved):
I think this looks fine - we'll refine when we know more.

PR #395, opened 2026-06-25
DRA-BUGFIX:  error #4 - Search was not properly reset when doing a new query after adding a TV/radio constraint.

jesperlauridsen:
Should do the trick


src/store/searchResultStore.ts
```diff
@@ -131,7 +131,9 @@ export const useSearchResultStore = defineStore('searchResults', () => {
 				});
 			} else {
 				const str = URLFilters;
+				console.log('here', URLFilters);
```
Pbruun: Console log ;)

src/store/searchResultStore.ts
```diff
@@ -131,7 +131,9 @@ export const useSearchResultStore = defineStore('searchResults', () => {
...
 				if (str.split('%3A')[0].includes('origin')) {
+					console.log('hullo');
```
Pbruun: Console Log ;)

Pbruun (changes requested):
Just some console logs. Otherwise it looks great

Pbruun:
No more console.logs so I'll merge :smile:

PR #397, opened 2026-06-25
DRA-BUGFIX: no desc or title search on star searches

jesperlauridsen:
Test number #4. This should make sure this issue doesn't arise again as we lock the buttons when there is no data to actually filter for title or description on.


Pbruun (approved):
I think this is a good way to solve the issue. I'll merge

PR #399, opened 2026-06-26
DRA OPTIONAL variable color

Pbruun:
We can use var instead of hexcoded color for custom components, this makes sure we use that instead.


jesperlauridsen (approved):
Very slick - good work. Works wonders for our approach to streamline the color codes! Good job!

PR #405, opened 2026-07-01
Dra 2762 spots

Pbruun:
We have yet to have the correct text we wish for these spots, so this PR is the first runthrough of spot integration


jesperlauridsen (approved):
Looks good.

PR #407, opened 2026-07-01
Dra 2789 searchbar redesign

Pbruun:
Searchbar styling. Includes hero branch so review this after.


jesperlauridsen (approved):
Good start to it. We'll tinker with the animation and placement later

PR #408, opened 2026-07-02
DRA-ZXYA: Remove icons and use material icons

jesperlauridsen:
First stab - removes the old icons and uses the material-icons instead, with the possbility of using outlined/filled.


Pbruun:
Looks fine, I didn't find any issues. The giant switch for getting the icon names is rough, however cannot come up with a better solution. I'll approve and merge when the other PRs are approved.

Pbruun:
Great work! I'll merge

PR #410, opened 2026-07-06
Dra scroll gridresults frontpage

Pbruun:
We wish to have slider view for the result items on the frontpage. So it takes up less vertical space.
This reuses the ItemSlider component and makes the adjustments so its usable.
Reason to reuse the component is to keep the same scoll speed logic throughout the site.


jesperlauridsen (approved):
I like this alot. A few minor details missing, layout wise, but we'll do those later i think!

PR #416, opened 2026-07-09
Dra optional period search no dropdown

Pbruun:
Search by period has been removed from custom components. is now fully visible throughout breakpoints.
Made some changes for the slider placement.


jesperlauridsen (approved):
Looks good.

PR #417, opened 2026-07-13
Color change. Use icon instead of background

Pbruun:
Search via period dropdown style changes. Unable to change the style of the options, but it doesn't appear we got any changes there on production either.


jesperlauridsen (approved):
Looks good.

PR #418, opened 2026-07-13
DRA 2839 result item colors

Pbruun:
Runthrough of the colors for the list search view. The thumbnails used the same class as other transparent dropdowns, but they weren't in UX design, so maybe they're no longer needed? So I outcommented, otherwise we need another class for KBButton.


Pbruun:
DRA-2836 was mistakenly added to this PR. Reason why the scope of this PR has expanded :sweat_smile:

jesperlauridsen (approved):
Lotta css color changes. Looks good.

PR #420, opened 2026-07-13
DRA 2841 program placement

Pbruun:
Removed margins so program view is directly below headers 2 lines. removed header heights, and changed breadcrumb top to margin. This was in order to have a better view of the program's placement.
Removed the background color so its white according to design.


jesperlauridsen (approved):
Good work.

PR #421, opened 2026-07-13
Dra 2831 breadcrumb adjustment

Pbruun:
merged PR https://github.com/kb-dk/ds-web/pull/420 into this, so take that first.
Otherwise the PR title speaks for it self.


jesperlauridsen (approved):
Looks good - according to design. I've ripped out all the remaining functions that were no longer in use.

PR #423, opened 2026-07-13
DRA 2863 searchbar full width

Pbruun:
When on the search page, the searchbar was not full width, but it was because I used Firefox. width: stretch is not supported on Firefox and Safari it seems. But they have their own version of stretch, so I added those.


jesperlauridsen (approved):
Good find. We gotta test this in safari/mac land aswell

PR #425, opened 2026-07-13
Added border around autocomplete. removed redundant code.

Pbruun:
Border around autocomplete. Code cleanup.
Still some issue with border-radius around the search-box. It seems to fade around the corners. Couldn't find a quick fix


jesperlauridsen (approved):
Good cleanup and finessing

PR #422, opened 2026-07-13
Dra 2837 sort buttons color

Pbruun:
Branched from #418
Simple PR where title says it all

PR #429, opened 2026-07-16
Dra 2833 button colors

Pbruun:
Pulled PRs into this: #420 #427 #422 
Most KBButton color were correct, but Unknown and NotAvailable didn't have KBButtons yet. So this PR became getting those sites adjusted. 
btn-tag colors were off, so they should be according to design now. 
On ShowRecord we had min-height 100vh. This messed with Unkown and NotAvailable, because footer got to be too far down.

A bunch of css cleanup on the components I worked on.


jesperlauridsen (approved):
Good work aligning these!

PR #431, opened 2026-07-17
DRA 2874 spots changes

Pbruun:
Its been decided that the info spot will be removed, and the read more spot will be larger.
Read more gained some more text, so I've added localization for that text and for categories title.

Cannot remember what was decided about the spot height. The read more has gained more content, and thereby more height, especially with English text. 
The English translation was done by me, has not been through UX or PO.


jesperlauridsen (approved):
Looks fine!

PR #432, opened 2026-08-05
Add Docker Compose support and hot reloading

seskildsen:
Add:
- Docker Compose support.
  - hot reloading inside container when changing code from outside.
  - ds-web container can talk with containers spun up from ds-backend (https://github.com/kb-dk/ds-backend) for local development.


README.md
```diff
@@ -39,3 +39,41 @@ npm run lint
...
 See [Configuration Reference](https://cli.vuejs.org/config/).
+
+# Local development with Docker Compose and hot reload
+
+Clone frontend `ds-web` repository and then clone `aegis` repository inside `ds-web` to get the `.env` used in `Docker Compose`:
+
+```shell
+git clone git@github.com:kb-dk/ds-web.git
+cd ds-web
+git clone git@github.com:kb-dk/aegis.git
```
jesperlauridsen, line 50: I don't like the idea of cloning the entire aegis repo into our repo in order to get it running in docker. We just need that one .env file for the repo to run, so a check for aegis somewhere else, and then getting the file, would probably be a safer and better way here, imho.

package.json
```diff
@@ -7,7 +7,7 @@
     "node": ">=20.17.0"
   },
   "scripts": {
-    "dev": "vite",
+    "dev": "vite --host",
```
jesperlauridsen, line 10: We should probably have a new "dev-docker", so it still works as it did before, and we don't hijack the current process.

jesperlauridsen (changes requested):
Just a few comments to look over - I don't like the idea of the entire aegis project lying around inside the repo here.

PR #434, opened 2026-08-07
DRA-2832: animations

jesperlauridsen:
Made transitions between pages and the logo more smooth and consistent.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -489,6 +489,10 @@ export default defineComponent({
 	transition: all 0.5s ease-in-out 0s;
 }
 
+.content.record {
+	margin-top: 20px;
+}
+
 .search-to-home-leave-to {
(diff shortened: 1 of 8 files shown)


src/App.vue
```diff
@@ -609,9 +645,13 @@ body {
 	overflow-x: hidden;
 }
 
+footer {
+	flex-shrink: 0;
+}
+
 .content {
+	min-height: 100vh;
```
Pbruun: The /find page with no query will have too much whitespace with this. Same with 404 page

Pbruun:
The 404 page has the same logo animation as the frontpage. I think it should be the same as search instead.

Pbruun (changes requested):
I see no issues with the code. Just had a few comments that needs to be looked at before I'll merge 😃

PR #438, opened 2026-08-07
DRA-OPTIONAL: Filters menu restyled

jesperlauridsen:
Refactor styles for radio buttons, checkboxes, and facets: update colors to use new CSS variables


Pbruun (approved):
Looks good, I'll merge. We'll take a talk with UX regarding header color after

PR #440, opened 2026-08-10
DRA OPTIONAL hover color for timeline

Pbruun:
The buttons on the timeline had old color scheme for hover effect. We do not have design for the new hover color, but it follows other button hover color design


jesperlauridsen (approved):
Looks good!

